PR #1, opened 2025-01-14
Add a readme

m6urns:


m6urns:
Screenshot from 2025-01-17 15-42-54

m6urns:
### Assignment 2 - Improved Tokenizer

The improved tokenizer builds on the simple tokenizer provided to us at the beginning of the assignment. To improve the performance of the tokenizer we implemented two key modifications: a frequency-based vocabulary selection and basic subword tokenization. Evaluation of the provided simple tokenizer and the improved tokenizer indicated a ~4% improvement in model performance when utilizing the improved tokenizer. The improved tokenizer achieves these performance increases by tracking token frequencies and maintaining only the most commonly (n=1000) used tokens. This approach is similar to a feature commonly implemented in the BPE tokenizer. 

The second and most significant improvement came from the addition of subword tokenization. When encountering unknown words, rather than immediately replacing them with an UNK token, the improved tokenizer attempts to break them down into meaningful subcomponents (prefixes and suffixes) that exist in its vocabulary. This is accomplished by identifying common word parts that appear frequently in the training data and using them to decompose unknown words. For example, if the word "lightsaber" isn't in the vocabulary but "light" and "saber" are common subwords, the tokenizer can represent it using these components rather than marking it as unknown. Implementing subword tokenization improved the model predictions by ~2% over the intermediate tokenizer with only frequency-based vocabulary improvements. Implementing both frequency-based vocabulary and subword tokenization improved tokenization performance and lead to a ~4% increase on a basic classification task using the provided simple tokenizer and the improved tokenizer.

__pycache__/bpe_sw_tokenizer.cpython-312.pyc
msharp9, line 1: As a note, __pycache__ files are great to include in your .gitignore file.

improved_sw_tokenizer.py
```diff
@@ -0,0 +1,190 @@
+import re
+from collections import Counter
+from typing import List, Dict, Set
+from zenml import pipeline, step
+from datasets import load_dataset
+
+class SimpleTokenizer:
```
IMJONEZZ, line 7: I am loving reading through your improved tokenizer, this is exactly the stuff I wanted to see from this assignment. You copied the simple tokenizer and the BPE from OpenAI, and then you implemented your own. Perfect!

test_tokenizers.py
```diff
@@ -0,0 +1,128 @@
+import numpy as np
+import re
+from sklearn.model_selection import train_test_split
+from sklearn.linear_model import LogisticRegression
```
IMJONEZZ, line 4: Amazing test file, happy to see you comparing them. It'll be great to see them compared using non-linear models to see the performance diff.

test_tokenizers.py
```diff
@@ -0,0 +1,128 @@
...
+            return tokens[:max_length]
+        return tokens + [tokenizer.str_to_int['<|endoftext|>']] * (max_length - len(tokens))
+    
+    # Create feature vectors
+    X = np.array([text_to_feature_vector(text, tokenizer) for text in texts])
+    y = np.array([char_to_id[label] for label in labels])
+    
+    return train_test_split(X, y, test_size=0.2, random_state=42)
+
+def evaluate_tokenizer(tokenizer, name="Unnamed Tokenizer", vocab_size=1000):
```
msharp9, line 54: This is really great work. I love that you created tests to determine which tokenizer is better.  This shows us your approach and thinking and it's spot on.  Great job you should be proud of this assignment.

As a note, a logistic regression model is likely not the best way to evaluate if the tokenizer is working. A better approach would be to find a measure to evaluate the tokenizer directly, like sending documents to be encoded and decoded and counting the number of unknown tokens.

PR #3, opened 2025-01-14
My first commit

matthewwhite1:
Assignment 2 Question 1


matthewwhite1:
image

matthewwhite1:
image

README.txt
```diff
@@ -0,0 +1 @@
+yeah
```
IMJONEZZ, line 1: This should probably be a README.md, just a note

run.py
```diff
@@ -0,0 +1,103 @@
...
+   def decode(self, ids): 
+      text = " ".join([self.int_to_str[i] for i in ids]) 
+      text = re.sub(r'\s+([,.:;?!"()\'])', r'\1', text) 
+      return text
+ 
+
+
+@step
+
+def tokenizer(vocab: list, text: str) -> list:
```
IMJONEZZ, line 44: I think I did a triple take at this implementation when I saw it. Crazy to be instantiating your class inside a step function, but if it works it works!¯\_(ツ)_/¯

run.py
```diff
@@ -0,0 +1,103 @@
...
+               count += 1
+      pairs1 = [item if item in self.str_to_int  
+            else "<|unk|>" for item in pairs] 
+      ids = [self.str_to_int[s] for s in pairs1]
+      pairs2 = [item if item in self.str_to_freq  
+            else "<|unk|>" for item in pairs] 
+      freqs = [self.str_to_freq[s] for s in pairs2]
+      return [ids, freqs]
+
+   def decode(self, ids):
```
IMJONEZZ, line 35: Looks like this outputs every character with a space in between them, none of the original spaces preserved.

PR #7, opened 2025-01-16
Jashon branch 2

jashonnew:
Jashon first commit


msharp9:
Great work!!  Just as a note, you mistakenly committed your entire python environment, so it was very difficult to find your work amongst the 800+ files you added here.

Your code could use some improvements when it comes to Python and OOP, but it looks good.

In your explanation for question 4, we would love for you to go more in depth into your thought process and your approach overall.

Please reach out for some office hours so we can help fix any environment issues you may still be running into.

IMJONEZZ:
I just really want to emphasize and acknowledge the amount of work you put in here. Great work, let's work together and figure out how we can make it easier for you down the line.

PR #8, opened 2025-01-16
Konner brown 2

brownkon:


starWars.py
```diff
@@ -0,0 +1,55 @@
+import json
+import re
+
+def get_pairs(word):
```
msharp9, line 4: Great job including this, looks like you had good ambition, however, we don't see that this is called anywhere.

Explanation.txt
```diff
@@ -0,0 +1,4 @@
+First off I stole the get_pairs function from the ChatGPT 2.0 code.
+This helps to understand more complex words with suffixes and prefixes such as unbelievable or greatly.
+I also included ". . ." as one of the symbols because that has a different meaning as just "." and should be its own token.
+I also exluded apostrophies from the symbols to split at to make contractions their own token instead of separate.
```
msharp9, line 4: This explanation is a great start. We would love to see a more in-depth solution.  We want to see your thoughts and approach.

starWars.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Return set of symbol pairs in a word.
```
IMJONEZZ, line 6: Great doc string, even if ChatGPT generated it. You should include these on your other functions later, code reviewers love these!

PR #13, opened 2025-01-28
Homework assignment 3

matthewwhite1:
Creating pull request!


matthewwhite1:
ChatGPT was incredibly helpful in helping me take the code I already had and putting it together in a way that worked for the data loader. Also, full disclosure: As I was reaching many hours on this assignment, I decided to just let ChatGPT write all of the tests for the data loader.

Here are the things my code does in order of appearance:

- Loads the Star Wars dialogue data
- Creates a vocabulary dictionary from this data
- Creates a tokenizer
- Creates a TextDataset using the sliding window approach
- Creates the data loader using PyTorch
- The pipeline function calls all of these steps to put it all together

This approach made a lot of sense to me. I originally had my steps in a very confusing order, but ChatGPT helped me realize that there was a better way to do it. If I had more time, I would make my tokenizer more complex, and write the tests myself.

main.py
```diff
@@ -1,2 +1,114 @@
...
+            return ids
+
+    tokenizer = SimpleTokenizer(vocab)
+    tokenized_text = []
+    for entry in data:
+        tokenized_text.extend(tokenizer.encode(entry["Line"]))
+    return tokenized_text
+
+
+class TextDataset(Dataset):
```
msharp9, line 73: Nit: This class is really a TokenDataset.  The tokens don't necessarily have to come from a text dataset. There's nothing wrong with this approach as tokenizing first can help speed up training down the road, but just for clarity.

tests/test_dataloader.py
```diff
@@ -0,0 +1,76 @@
+import pytest
+import torch
+from torch.utils.data import DataLoader
+from main import TextDataset, create_dataloader_v1, SimpleTokenizer
+import sys
+import os
+
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
msharp9, line 8: Since you already imported everything from main, I'm assuming this line isn't necessary.

msharp9 (commented):
Overall great work! 

A few notes though. Your CI/CD pipeline is failing. Be sure to format, lint, and ensure your tests are passing before pushing.  https://github.com/IMJONEZZ/USU-LLM-Class/actions/runs/13083340676/job/36510971472?pr=13

Also, you say if you had more time you would have written the tests yourself, why? The tests you have are probably better. What I would encourage you to do instead is think about what tests are missing, are there other edge cases not considered? That type of thing.

PR #16, opened 2025-01-31
HW assignment 3

rj-may:
This is my pull request for assignment 3


rj-may:
In a comment in your PR explain what your data loader does, how it works, and why you took the approach you did.  Explain why you wrote the tests you did.  If you had more time, what would you do to improve this PR? 

I have no clue how some of the data loader class stuff works. I just used the example code and got  into a working state.  I built something that combines the example code and plugging in the tokenizer into one working garbage heap that runs with ZenML. I think I am fighting ZenML more than anything else on these assignments. So the answer to the why I did what I did : find a way to beat ZenML. 
 Essentially the @step load_data uses the create_dataloader_v1 function builds a tokenizer and uses that to create the TextDataset object.  The approach was ChatGPT, DayQuil combo of awesomeness. 


If I had more time to improve this stuff, I would figure out how to use my @load_data step to do something with it (like training). There is always an issues when I try and pass it to another step. ZenML complains a bunch about it being serialized, pickled or not the right file dtype when I can use the functions when I am working on a scratch file like a notebook. 

The tests:
Well I wasn't sure how much information could be extracted from the dataloader object, so I lot of discussion with chat gpt  we got a  couple of things. We tested a different batch sizes (like 1), tested if it was going multiple epochs, and I tested if it was an instance of a dataloader.  Also, the dataloader tested different things sucha s the shape of the batch and the lengths. Those all seemed like good things to look at.  Also, with chat we tested if the TextDataset set was working with right data types (torch tensors). 



Also, a note on torch. I don't have NVIDIA GPUs, so I downloaded the cpu version of pytorch. I am not sure if that is slowing me down at all.

.github/workflows/ci.yml
```diff
@@ -26,3 +26,5 @@ jobs:
     - name: Run tests
       run: |
         uv run pytest
+    - name: Create mock file
+      run: echo "This is a test file." > mock_file.txt
```
msharp9, line 30: I have no idea what you need this.

tests/.mock_file.txt.swp
msharp9, line 1: I'm assuming this file should be git ignored.

tests/test_dataloader.py
```diff
@@ -0,0 +1,144 @@
+import pytest
+import torch
+from torch.utils.data import DataLoader
+
+import sys
+import os
+
+# Add the parent directory to the Python path
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
msharp9, line 9: This shouldn't be needed, see the test_main.py.  Is this a Windows thing maybe?

tests/test_dataloader.py
```diff
@@ -0,0 +1,144 @@
...
+def get_string(file_path):
+    return "Lorem ipsum dolor sit amet."
+
+
+# # Test the load_data step
+def test_load_data():
+    # file_path = "mock_file.txt"
+    # dataloader = load_data(file_path)
+
+    with open("mock_file.txt", "r") as file:
```
msharp9, line 77: Ah, I bet this was likely confusing to you.  Since pytest will be run from the root directory.  It also explains the two mock_file.txt files.  You should point to `"tests/mock_file.txt"`

msharp9 (commented):
Great work over all!  Sorry to hear you were sick. Hope your need for DayQuil is already dissipating.

The CPU version is likely not slowing you down at this point. There are several resources where you can use a free GPU online, the most common one being Google Colab.

PR #17, opened 2025-01-31
Homework Assignment 3

jashonnew:


jashonnew:
ChatGPT helped massively. It also helped quite a bit to take a step back and review some basics (turns out a pull request has nothing to do with pulling the repo on my end LOL). I created a very basic data loader using the built in pytorch class, as there was no direction towards complexity. I know that the dataloader is an absolute NECESSITY for LLMs because the lazy data style or reading on these massive datasets (discussed in chapter 3) such as the full web in strings and such is too big for your pc to read in. you have to batch read and load into your tokenizer in bite sized pieces. 

The testing portion is almost entirely suggested by chatGPT however the reasoning behind the testing is solid. We test for the following reasons:
1. Basic Functionality -> does it batch load data
2. Ensure correct batching and batch size
3. Ensure data is shuffled properly 
4. Handling edge cases
5. handling an empty data set


Unfortunately there is some sort of bug coming from a pytorch zenml interaction, but at 10 hours invested in learning about LLMs and code implementation I have to throw in the towel for the week.

.gitignore
```diff
@@ -0,0 +1,55 @@
...
+
+# translation temp files
+po/*~
+
+# RStudio Connect folder
+rsconnect/
+
+# Common Data Types
+.data
+.csv
```
msharp9, line 55: This isn't a big deal so don't stress on this, but a note on keeping your git history clean.

This looks like an R based .gitignore file, because of this, you are including a bunch of files in your commit I bet you don't intend to, like .ipynb_checkpoints, __pycache__, and others.

run.py
```diff
@@ -0,0 +1,62 @@
+from zenml import pipeline, step
+import json
+import re
+import pandas as pd
+from typing import List
+from collections import Counter, defaultdict
+from torch.utils.data import Dataset, DataLoader
+from BPEtokenizer import BPETokenizer
+from DataLoader import DataLoader
```
msharp9, line 9: This is just re-importing the torch DataLoader you have on line 7.  I believe what you meant to import was the CustomDataset.

run.py
```diff
@@ -0,0 +1,62 @@
...
+    x = dummy_step()
+    
+    
+    #Read in
+    with open('SW_EpisodeIV_VI.json', 'r') as file:
+        raw_data = json.load(file) 
+    if isinstance(raw_data, dict):  # If JSON is a dictionary, extract string values
+        data = [str(value) for value in raw_data.values()]
+    elif isinstance(raw_data, list):  # If JSON is a list, convert all items to strings
+        data = [str(item) for item in raw_data]
```
msharp9, line 49: This type of logic might make more sense to put in the Dataset itself, and just pass in the raw_data.

run.py
```diff
@@ -0,0 +1,62 @@
...
+    else:
+        raise ValueError("Unexpected JSON format: Expected a dictionary or list.")
+    dataset = load_data(data)
+    # Initialize DataLoader
+    dataloader = DataLoader(dataset, batch_size=8, shuffle=True, num_workers=0)
```
msharp9, line 54: So this line would best fit in the load_data step.

run.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pipeline
+def llm_pipeline():
+    """Define a pipeline that connects the steps."""
+    x = dummy_step()
+    
+    
+    #Read in
+    with open('SW_EpisodeIV_VI.json', 'r') as file:
```
msharp9, line 44: So, ideally a pipeline isn't running any code itself and is just kicking off steps.  So this might be best to put in a @step called read_json_file or something similar.

msharp9 (commented):
Excellent work. A few notes to help you along.

It looks like you forgot to run `git pull` before creating your branch so it isn't based off the latest main commit. This may have made things more difficult, but also because of this your CI/CD pipeline didn't even run because it doesn't exist in your branch. Unfortunately, getting this pipeline passing was part of the assignment of writing the tests to show that they are passing. 

I do love that you used ChatGPT to help you write the tests though.  Great place to save you time.

PR #19, opened 2025-01-31
Homework assignment 3

tjkerby:
Initial pull request for homework 3


tjkerby:
My dataloader initializes by loading in the Star Wars scripts for episodes 4-6 from a file_path to the json file. It also has the required __len__ and __getitem__ functions. The dataset is composed of observations  that consist of a dict with the 'Character" and 'Line' keys. So a single observation would look like {'Character': 'THREEPIO', 'Line': "We're doomed!"}. Dataset.__getitem__(1) would return the second line in the movie scripts and dataset.__len__() would return the total number of lines in the movies (in this case, 2523). 

I chose to run 3 tests. The first makes sure that the len attribute works on a sample dataloader. The second ensure that the data is stored in the expected format. The third makes sure that you can properly iterate through the dataloader without issue. If I had more time, I would likely like to improve the sample dataloader as it is pretty simple and short at this time. Additionally, it would be nice to test things like ensuring that it handles edge cases nicely like when a line is blank or things of that nature.

tests/test_dataloader.py
```diff
@@ -0,0 +1,55 @@
+# test_dataloader.py
+import json
+import pytest
+from torch.utils.data import DataLoader
+import sys
+
+sys.path.append("C:/Users/tjker/Desktop/School/Spring_2025/USU-LLM-Class/usu_llm_class")
```
msharp9, line 7: You shouldn't need this right? It's passing in the CI/CD pipeline which I'm sure doesn't have this path.  Is this a Windows thing?

msharp9 (commented):
Good job.  I like how you kept things simple and very clean.  Be thinking about how you would integrate your tokenizer into your dataloader.

As a note, keeping a dataset in tests very small is generally a good thing to help them run fast, however, it should be large enough to test all edge cases.  Using a Copilot to help you write tests is encouraged to help you get more coverage for your code.

PR #21, opened 2025-02-01
Nathan weber 3

Nathan-Weber208:
I was getting tired of running all the command to restart the dockerfile so i made restart_docker.bat
If there is a way to change run.py while zenml is up and running id love to hear about it. I hate having to do the signup process every time i want to test my functions.

Question 4:
The data loader basically makes the data stream into learnable batches. 
mine will:
Accept a batch size so it will fit the trainer.
Shuffle the data to avoid overfitting.
Control number of parallel processes are running.
Use Pin memory.

Those last 2 were suggestions from chat GPT. apparently pin memory will allow it to be trained using the GPU.

a function i added is collate_fn
this will normalise the inputs so that it all is fed at the same shape.

I asked chat GPT to write my tests for me. specifically with collate_fn i really didn't know what to expect as an input.
I got them all to pass.

If i had more time, I think i would explore pin memory more, try and run it on my local machine. I have a pretty good graphics card for a laptop so id like to give it a whirl.


Nathan-Weber208:
I finally got it to pass. i might need to meet with you just to see how you set it all up

.github/workflows/ci.yml
```diff
@@ -2,27 +2,51 @@ name: Test, Format, and Lint
...
+      - name: Verify Installed Packages
+        run: |
+          uv run python -m pip list
+          uv run python -c "import torch; print('PyTorch installed:', torch.__version__)"
+
+      # 6) Set PYTHONPATH and Run lint checks, format checks, tests
+      - name: Run Lint, Format, and Tests
+        run: |
+          # Add project root to PYTHONPATH so 'run.py' is importable
+          export PYTHONPATH=$PWD:$PYTHONPATH
```
msharp9, line 43: Interesting that this was required for you. If you compare it to test_main, you'll see it it's able to find main.py and load in hello_world on the main branch.

dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN pip install --no-cache-dir "zenml[server]"
+
+# Initialize ZenML
+RUN yes | zenml init
+
+# Expose the ZenML server port
+EXPOSE 8080
+
+# Copy required files (but NOT pytorch_packages)
+COPY run.py /app/
```
msharp9, line 16: So, it looks like you are running things slightly different from others. You should be able to just kick off the docker image to start a local ZenML server, then use zenml login to connect to it.

One trick for your current set up that might help.  You can add your directory as a volume, this way local edits will show up in your docker container.  You do this when creating the container with docker run.

Something like (it'll be different on windows):
`docker run -v ${PWD}:/app ...`

This will replace the contents of /app/ in the docker image with the contents of your present working directory.

msharp9 (commented):
Great work.  I loved what you did.

Looks like you ran into some issues, but got it working in the end.  Good job persevering.

PR #22, opened 2025-02-01
Elias thompson 3

eliasthompson304:
For some reason putting test_dataloader.py in a separate "tests" directory made it difficult for me to import my dataloader module and run tests. Thus, my test_dataloader.py is in my src directory along with my dataloader and pipeline script. My pipeline script is still named tokenizer.py, which I should change soon. 

How I came up with my DataLoader:
After glancing around pytorch's website and finding their built in dataloader, I decided to give it a shot for my pipeline. The code resembled something like this:

from torch.utils.data import DataLoader
train_dataloader = DataLoader(training_data, batch_size=64, shuffle=True)

This worked just fine for my dataset, tokenizer, and zenml pipeline. It broke down our star wars dataset into batches containing unique inputs and labels. However, this felt too easy for the assignment. I decided to get some help from AI to make a similarly functioning DataLoader on my own.

How it works:
My CustomDataLoader takes a TensorDataset (pytorch imported) and applies it to our dataset that we've prepared and organized into inputs (lines of dialogue) and labels (characters). Then, we assign the dataloader class a batch size attribute, so we can later divide the dataset into batches for training and testing. We can get our batches by iterating through our dataset in chunks the size of "batch_size." We also make a list of indices that contains index positions of items in the dataset. Now, because each index of the dataset is a tuple containing input and a label, we can retrieve batch data by using a list comprehension and our batch_indices.  After converting the batch data into tensors just like pytorch's default DataLoader, we can get the number of batches using the __len__ method.


eliasthompson304:
for some reason my files in the build checks (that failed) do not seem to match the latest commits I made attempting to fix them.

eliasthompson304:
Screenshot 2025-02-01 at 6 20 33 PM
My pipeline and pytests run just fine, but I can't seem to format them properly. This is error I get when I push. Any advice on what my next step should be?

msharp9 (commented):
Overall impressive work.

First off, I love your custom data loader.  But to answer your question, it was just that easy. We actually expected you to create a custom dataset, not a custom dataloader.  A lot of the code around loading, tokenizing, and turning into a TensorDataset will normally happen there, in a CustomDataset.

As far as getting ruff to pass. `ruff` is both a linter and a formatter.  Formatters actually fix your code for you so you just needed to run `uv run ruff format` as it says in the Readme.  I hope that helps for next time.

eliasthompson304:
thanks!

PR #23, opened 2025-02-05
Homework assignment 3

eliasthompson304:
Homework Assignment 3 Question 1


eliasthompson304:
When using the terminal, my tests both pass. Yet even with a requirements.txt file in my root dir and appropriate changes made to my .yml file I still get an error in the building phase. Error: no module named 'numpy.'

Anyways here is what I think about my evaluator and model:
My model did ok. I split my data into three categories: a training set, testing set, and validation set. Split up into 5 epochs I got on one run 20% accuracy (yikes). As of now my model takes encoded tensors of dialouge lines and attempts to predict which character (or label) it belongs to. Here is an output demo:

Screenshot 2025-02-07 at 9 29 58 PM

I will need to find ways to improve my model and likely represent the data better if I want its accuracy to increase. Right now I have it set up to 60% training, 20% validation and 20% testing. Should I alter this ratio?

For my first test, I wanted to ensure that my model was splitting data correctly. I did this by checking if there were overlaps between training, validation, and testing data. If so, I wanted to print out how many. For my next test, I used torch (just like in my model) to evaluate whether the accuracy of my model was being reported correctly.

I did indeed format my code with ruff. My pipeline and tests run after formatting yet still don't pass the checks on GitHub. This is my first time dealing with these checks but I'm a little annoyed I haven't figured them out yet.

eliasthompson304:
This is how my evaluator works:

* I keep track of # of correct predictions
* keep track of total predictions made
* with torch.no_grad() is a apparently a trick that saves memory and computational resources (AI assisted)
* iterate through data from test_dataloader (dataloader for our test dataset) in terms of batch_inputs and batch_labels (inputs = tokenized lines of dialogue, tests = character labels).
* pass batch_inputs through our trained model (a text classifier that uses torch.nn
*  add number of samples in current batch to a total 
* add correct predictions to var correct
* compare predicted labels with the actual labels
* .sum() counts instances of True
* .item() gets number of correct predictions
* calculate accuracy using maths (100 * correct /. total)
* print test accuracy result

This was my first time training and evaluating a model, so this approach was fueled mainly by a desire to get it working. I look forward to fine tuning it more over the weekend and seeking to understand other possibilities.

.github/workflows/ci.yml
```diff
@@ -14,10 +14,12 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v5
       with:
-        python-version-file: ".python-version"
+        python-version: "3.12.7"
 
     - name: Install dependencies
       run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
```
msharp9, line 22: This isn't going to help you since pip will install it globally, but uv runs in a virtual environment.  Since we run the tests inside the virtual environment with `uv run pytest`.  In the README it explains to add dependencies simply run `uv add cowsay` or in your case `uv add numpy`.  This will add it to both the pyproject.toml as well as the uv.lock files.

For additional clarification, uv uses uv.lock by default since they provide some additional features.  If you'd like to use a requirements file with uv, you could change this line to `uv pip install -r requirements.txt`.

You can read more here: https://docs.astral.sh/uv/concepts/projects/dependencies/

.gitignore
```diff
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+#dataset
+src/starwarsdataset.json
+src/starwardataset.json
```
msharp9, line 8: It's good to add your datasets to the ignore list.  Unfortunately you had already added your datasets before adding them to .gitignore.  .gitignore will only ignore untracked files.

See here for details on how to fix it: https://stackoverflow.com/questions/1274057/how-do-i-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore

README.md
```diff
@@ -1,5 +1,5 @@
 # USU LLM Class
-
+#change made for question1 assn3
```
msharp9, line 2: Assn4 :)

tests/test_evaluator.py
```diff
@@ -0,0 +1,42 @@
...
+
+    # all possibilities for an overlap
+    train_val_overlap = train_inputs & validation_inputs
+    train_test_overlap = train_inputs & test_inputs
+    val_test_overlap = validation_inputs & test_inputs
+
+    # prints, if any, overlaps in train, test, or validation datasets (ensure data splitting happened correctly)
+    print(f"Train-Validation Overlaps: {len(train_val_overlap)}")
+    print(f"Train-Test Overlaps: {len(train_test_overlap)}")
+    print(f"Validation-Test Overlaps: {len(val_test_overlap)}")
```
msharp9, line 32: Just a note, in tests, print statements aren't very meaningful.  Figure out how to use an assert statement like you do below.

tests/test_evaluator.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+# since the predictions match the labels this should test if my model's accuracy reports work (I should get true on 100% accuracy)
+def test_accuracy_of_eval():
+    predictions = torch.tensor([0, 1, 2, 3])
+    labels = torch.tensor([0, 1, 2, 3])
+    correct = (predictions == labels).sum().item()
+    total = labels.size(0)
+    accuracy = 100 * correct / total
+    assert accuracy >= 100.0
```
msharp9, line 42: This probably doesn't need to a be `>=` since this test is hard coded.  Which also means, unfortunately, you are not actually testing your evaluator anywhere.  Similar to your code above where you import and use the tokenize_data method, you'll want to do the same here.

It's usually easier to test smaller methods and functions vs big ones, so don't be afraid to break things up.

msharp9 (commented):
Hey, this is really great work. Good job getting the training loop to work. Your pipeline: load, train, evaluate is perfect. Looks like you just had some hiccups with using `uv` that prevented you from getting your tests to pass.  Glad to see you got `ruff` to work this time though!

Overall well done and well written explanation.  For your question on ratio's, it really depends on your dataset.  You need a large enough training data for your model to learn, so on a smaller dataset like you are using I might increase it's percentage. You validation dataset needs to be large enough to ensure a diversity comparison to the training dataset so you can be confident you are generalizing well. Testing is usually the smallest dataset of the three, but giving it 20% on an assignment to create an evaluator makes sense.

PR #24, opened 2025-02-06
Homework assignment 4

matthewwhite1:
Starting assignment 4!


matthewwhite1:
I will be transparent again and say that ChatGPT did much of the heavy lifting here. Even with its help, I still encountered so many errors that made this whole process pretty frustrating.

One common error I've been encountering for the last few assignments is having trouble accessing ZenML output. Things being stored in stuff like StepArtifacts and however it stores its pipeline information has made accessing this data very confusing.

Also, I fixed some of the more basic CI pipeline errors, but the most recent one is one I don't understand how to fix. It seems like it doesn't recognize my assignment_4_pipeline as a recent pipeline run, even though it is the most recent pipeline run in my ZenML client.

The evaluator uses a cross-entropy approach to compute loss. I don't understand entirely what it's doing in my code, but it makes sense as a way to compute loss.

Just like last week's assignment, ChatGPT wrote all of the tests.

If I had more time, I would look deeper into ZenML and PyTorch documentation to better understand many of the errors I was getting. I think I should probably schedule an office hours appointment next week to work out some of my confusions.

evaluator.py
```diff
@@ -0,0 +1,59 @@
...
+from torch.utils.data import DataLoader
+from zenml.client import Client
+
+# Load tokenizer and model
+tokenizer = BertTokenizerFast.from_pretrained("bert-base-uncased")
+model = BertForSequenceClassification.from_pretrained("bert-base-uncased")
+model.eval()
+
+
+def evaluate_model(dataloader, model):
```
msharp9, line 12: You probably would have saved yourself a lot of trouble if you simply made this another step in your pipeline.  You might have wanted to avoid this so you didn't have to run the entire pipeline while debugging. Tokenizing and create your dataset each time could take a long time after all, but you could have simply paired down you SW_EpisodeIV_VI.json to a much smaller dataset for testing.

evaluator.py
```diff
@@ -0,0 +1,59 @@
...
+            # Assuming single-class problem (adjust accordingly if needed)
+            labels = torch.zeros(outputs.shape[0], dtype=torch.long).to(device)
+            loss = torch.nn.functional.cross_entropy(outputs, labels)
+
+            total_loss += loss.item()
+
+    return total_loss / len(dataloader)
+
+
+# Initialize ZenML client
```
msharp9, line 35: As a note, everything below this should be in a `if __name__ == "__main__":` block.  All of this code is trying to run when you import the file in your tests on line 4 `from evaluator import evaluate_model`.

This is why your tests are failing, because of course github action runners don't have access to your ZenML running locally on your computer.

evaluator.py
```diff
@@ -0,0 +1,59 @@
+from transformers import BertTokenizerFast, BertForSequenceClassification
+import torch
+from torch.utils.data import DataLoader
+from zenml.client import Client
+
+# Load tokenizer and model
+tokenizer = BertTokenizerFast.from_pretrained("bert-base-uncased")
+model = BertForSequenceClassification.from_pretrained("bert-base-uncased")
```
msharp9, line 8: You probably shouldn't but loading the tokenizer and model like this.  See my other comment about putting it in the main block.

msharp9 (commented):
This is good work. I'm sorry you got stuck, it happens to everyone.

It might help if you took a minute to read:
https://docs.zenml.io/user-guide/starter-guide/manage-artifacts
and 
https://docs.zenml.io/how-to/pipeline-development/build-pipelines/fetching-pipelines.

Getting a big picture overview may prevent some of your confusion. I'd do so outside of a homework assignment when you are trying to "do something". This can help prevent skimming over important information that doesn't seem relevant at the time but helps clarify how things work.

Unfortunately, your explanation of your evaluator is a bit lacking. The goal was to create a custom evaluator and there's lots of approaches you could have taken.  We'd rather you have a good plan over working code. To get full points next time, take some time to evaluate the pros and cons of your decision.  Using cross-entropy here is fine, but it appears you just chose the first thing ChatGPT gave you.  This might not be the case, but that's why we are asking you to explain your work.

Same thing for your explanation of your tests.  It's fine if ChatGPT wrote them for you, but it's still your job to evaluate if they are good. Since the pipeline is also failing, it suggests maybe they weren't.

PR #26, opened 2025-02-07
Homework assignment 4

rj-may:
I don''t see the testing pipeline ran on this page


rj-may:
My model evaluator did ok. I created a lemmatized TF-IDF that looks at the cosine similarity between the doc strings. I wish I would have done something slightly different. I wish I would have just lemmatized the sentences and looked at the proportion of words that were similar without stop words. I got a cosine similarity score of 0.4696 for validation and test score .5112 cosine similarity which means that is somewhat working. I was confused on how this assignment was supposed to work creating a validation and testing datasets. I literally just asked an LLM  (Gemini) to build my dataset. 

The tests that are here in github are only some of the tests that I was working. The tests that are in the working check some of the basic stuff. It check to see if the words got lemmatized, and the if the TFIDF score is in between 0 and 1.  Those 
I wanted to tests some other stuff such as testing some edge cases of getting completely different sentences, but I couldn't get that code to run for some weird reason. 


I am going to be honest and say that the majority of my code was written by ChatGPT. That's why my code is called chat2.py.  I wasn't able to see how to connect the dots between the code in the slides and the assignment description. I wasn't sure how to get test and validation sets (like where those coming from BertGeneration). Tying the pieces together somehow escaped making sense to me. So I guess I if had more time, I would get some help with this stuff. I would love to be able to write most of my own code, using chatgpt is not really my style. I want to understand how the pieces are coming together. On the flip side of the coin,
 I just can't dedicate tens of hours to this stuff between thesis writing and job applying.

chat2.py
```diff
@@ -0,0 +1,372 @@
...
+    # Validation
+    print("\nValidation Evaluation:")
+    val_score = evaluate_tfidf(model, val_dataloader, device)
+    print(f"TF-IDF Cosine Similarity Score: {val_score:.4f}")
+
+# Step 8: Save Model
+torch.save(model.state_dict(), "bert_llm.pth")
+
+
+# Step 10: Final Testing
```
msharp9, line 356: Your steps got a bit out of order eh? 9 is up above between 6 and 7.

If you are creating steps with comments, why not just create them with ZenML?

tests/test_evalutor.py
```diff
@@ -0,0 +1,237 @@
...
+
+nltk.download("wordnet")
+nltk.download("omw-1.4")
+
+# Initialize lemmatizer
+lemmatizer = WordNetLemmatizer()
+
+
+# 1️ **Define the Model**
+class BERTModel(nn.Module):
```
msharp9, line 21: So the issue here, is that you aren't importing in your BERT_LLM class, you are just sort of recreating it.  In fact, you aren't importing anything from chat2.py so you aren't actually testing the code. Which unfortunately is an issue.  

Of course, importing chat2.py is going to cause problems since it's written as a script which will cause it to run, which is maybe why you just recreated it? You can either put all your code in methods, like a main() function (note: if it was ZenML code that would already be true), or you can be sure to keep all script related code in a `if __name__ == '__main__':` block

msharp9 (commented):
This is good work. I hope your commit comments are more facetious than literal.

PR #28, opened 2025-02-07
Homework Assignment 4

jashonnew:
-added assgn 4, which contains evaluation and training data

-created a testing script for the evaluator

-By recommendation of last weeks homework comments, I restarted from main. I am now curious about zenml, has that been abandoned? There is no longer a run.py folder. I have my work for previous weeks and even corrected my errors within zenml, but the absolute mountain of content we are pushing through each week makes me question if we are leaving that architecture behind?

- I didn't replicate my evaluation process with finetuning data, validation and test. While this would be more robust, I had some trouble wrapping my head around the evaluator in the first place and the integration of the evaluation within our current architecture.

- I would take more time to understand exactly imported functions and improve the training logic to include validation and testing. I would use text that interests me instead of generated evaluation training logic. All fluff had to be cut after 6+ hours on this assignment (my schedule is very packed).

### DIFF
--- a/SW_EpisodeIV_VI.json
+++ b/SW_EpisodeIV_VI.json
@@ -1,0 +1,10094 @@
+[
+    {
+        "Character": "THREEPIO",
+        "Line": "Did you hear that?  They've shut down the main reactor.  We'll be destroyed for sure.  This is madness!"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "We're doomed!"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "There'll be no escape for the Princess this time."
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "What's that?"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "I should have known better than to trust the logic of a half-sized thermocapsulary dehousing assister..."
+    },
+    {
+        "Character": "LUKE",
+        "Line": "Hurry up!  Come with me!  What are you waiting for?!  Get in gear!"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "Artoo! Artoo-Detoo, where are you?"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "At last!  Where have you been?"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "They're heading in this direction. What are we going to do?  We'll be sent to the spice mines of Kessel or smashed into who knows what!"
+    },
+    {
+        "Character": "THREEPIO",
+        "Line": "Wait a minute, where are you going?"
+    },
+    {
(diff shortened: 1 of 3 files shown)


SW_EpisodeIV_VI.json
msharp9, line 1: Make sure you add your dataset files to the .gitignore file.  We didn't explicitly call it out like we have in other assignments, but it should be done for each assignment.

tests/test_evaluator.py
```diff
@@ -0,0 +1,46 @@
+from nltk.stem import WordNetLemmatizer
```
msharp9, line 1: So your tests ended up failing on this line.  The commands you need to run are in the README for reference, but make sure you run `uv add nltk` to ensure that nltk library (or any other library you use) is added to the project. `uv add` will ensure that the library is added to both the pyproject.toml file and uv.lock file so that way it can be imported in the Github Action.

tests/test_evaluator.py
```diff
@@ -0,0 +1,46 @@
...
+    scores = rouge_scorer.RougeScorer(pred_lem, ref_lem)
+
+    # Display results
+    print(f"Test {i + 1}:")
+    print("Original Prediction: ", prediction)
+    print("Original Reference:  ", reference)
+    print("Lemmatized Prediction:", pred_lem)
+    print("Lemmatized Reference: ", ref_lem)
+    print("ROUGE Scores:", scores)
+    print("-" * 50)
```
msharp9, line 46: A pytest test will generally use an assert not a print statement.  A print statement will never fail.

So something like:
```python
assert prediction == reference
assert pred_lem == ref_lem
```

tests/test_evaluator.py
```diff
@@ -0,0 +1,46 @@
...
+# Ensure necessary NLTK resources are downloaded
+nltk.download("wordnet")
+
+# Initialize lemmatizer and ROUGE scorer
+lemmatizer = WordNetLemmatizer()
+rouge_scorer = rouge_scorer
+
+
+# Function to lemmatize a sentence
+def lemmatize_sentence(sentence):
```
msharp9, line 14: So the problem here, is that you are just recreating this method from your assgn4.py, which means you aren't actually testing the code.  What you should do is import it and then test it that way.

Of course, that's going to cause issues the way assgn4.py is written as a script. If you try to import it, it's going to run the entire file.  If your code was wrapped in ZenML steps that would circumvent the problem.  Alternatively you could put everything you don't want to export in a `if __name__ == '__main__':` block.

msharp9 (commented):
This code is clear and your direction is obvious, you did a good job on this assignment. Your python, git, etc. is getting better.  Just a little bit more effort on the tests.

To your question yes, while ZenML wasn't required on this assignment, we will still be using it. As we start approaching actually training and finetuning models next week, it will become more valuable to you.

PR #29, opened 2025-02-08
Homework Assignment 4

m6urns:
Q2: How well did your model do? Would you choose this model for your task? 

One of the advantages of this model was that is was trainable in only a small amount of time and with limited compute resources. However because of this the performance of the model is very limited. Looking at some basic metrics for the validation set below:

Loss: 6.3907, Perplexity: 596.2457
Token Accuracy: 1.0000

We can see a high degree of token accuracy and based on my understanding of perplexity this would indicate that the model is effectively memorizing/copying from the source rather than producing realistic dialogue. I would guess that this could be mitigated through longer training or on an augmented dataset. Or potentially by increasing the complexity of our model. Based only on the information from these tests I would not choose this exact model. 

Q4: Explain what you evaluator does, how it works, and why you took the approach you did. 

Some of the statistics reported by my evaluator have been previously provided. In addition to those statistics my evaluator also provides some basic evaluation of types of responses. 

truncation: 0.0%
repetition: 1.9%
generic: 17.2%
incoherent: 46.6% 

Looking at these results and some example of the kinds of dialogue produced by my model I believe that this backs up my previous claim that the model is good at producing tokens from the input. But is not sufficiently complex to produce actual believable dialogue. Some of the issues I observed included repetitive patterns in the text and the model appending irrelevant words to the ends of what would have otherwise been the beginning of good dialogue.  

I based my approach on the examples provided in the mlops starter and the hugging face BertGeneration documentation. I also needed to reduce the complexity of my initial model to allow for memory constraints. 

Explain why you wrote the tests you did. 

I created tests cases with the goal of validating each component in the evaluation step. The tests created basically look to see if evaluation metrics are being correctly calculated. In testing I did find an issue with the way I was calculating sequence accuracy and needed to go back and revise the way that it was calculated. 

If you had more time how would you improve this PR? 

Given more time to improve this PR I would write tests that could evaluate the integration of the model pipeline with the ZenML server.  

Screenshot from 2025-02-07 18-03-05


msharp9 (commented):
Wow, it appears you went above and beyond on this homework assignment. Well done!!!

PR #30, opened 2025-02-08
Tom kerby 4

tjkerby:
Homework 4 pull request


tjkerby:
I'm not going to lie, this was a rough one for me. I had a difficult time understanding how this homework built on from previous ones. In the end I used a bit more of chatgpt than I usually would have liked. As far as the model goes it does fairly poorly, but that isn't unexpected as I did not train it for long and BERTs training likely didn't help it much with generating cypher queries. I would definitely use a different model if I actually cared about the performance. I'm not sure if the tests are failing due to me not uploading the model.pt file, but it is too large to upload onto github so it is what it is. It passes on my computer. My evaluator.py test loads in the first observation from the text2cypher test dataset and makes the model predict the associated cypher query from the question. It then computes the BLEU score between the predicted cypher query and the actual cypher query. The model gets quite a low score as it learned that cypher queries tend to contain ":"  and parentheses. So it mostly predicts those tokens lol. This test tries to do everything at once, but if I had more time I would refactor the tests to test the data loading, then test to make sure that the model and tokenizer load and work properly, then test that the data is formatted as expected. Then test that the evaluator works as expected for normal cases and edge cases. I know this isn't my best work, but it ended up taking me far longer than I anticipated and as such I got to where I am.

tests/evaluator.py
msharp9, line 1: Your tests are failing because pytest couldn't find your tests.  It looks for files that start with `test_`, not necessarily files in a tests directory.

Reference: https://docs.pytest.org/en/stable/how-to/usage.html#how-to-invoke-pytest

tests/evaluator.py
```diff
@@ -0,0 +1,58 @@
+import torch
+import pytest
+from datasets import load_dataset
+from transformers import BertTokenizer
+import evaluate
```
msharp9, line 5: You aren't importing your `compute_metrics` method from your metrics.py file.  So unfortunately, you aren't actually testing your code at all.

msharp9 (commented):
Nice work! Seems like knowing a few HuggingFace tricks made the code much simpler for you.  Just some issues with the tests, but good job overall.

PR #33, opened 2025-02-10
Haozhe Chen HW4

chenhaozhe1:
In this homework assignment, I collected a set of reviews—both positive and negative—for the game Black Myth: Wukong and used my evaluator to predict their sentiment.

My evaluator utilizes a custom ReviewDataset class, which takes the game reviews and their corresponding labels, tokenizes them using a pretrained BERT tokenizer, and feeds them into the model in batches for evaluation. During this evaluation, the model processes batches of reviews, performs sentiment classification, and compares the predictions against the actual labels. The accuracy and classification report are then computed using sklearn.metrics. To prevent order bias during evaluation, the dataset is shuffled, ensuring that BERT does not inadvertently "learn" from the order of input data.

I wrote tests to verify several aspects of the evaluator:

Ensuring the ReviewDataset class correctly tokenizes the input reviews and aligns them with their corresponding labels. This is validated by checking the dataset’s length and retrieving individual items (including input_ids, attention_mask, and labels).
Verifying that the Evaluator is properly initialized with the correct model and tokenizer, ensuring the setup is correct.
Confirming that the evaluate() method returns the expected results (accuracy, classification report, and predictions), and that it classifies the reviews correctly, using a small dataset with both positive and negative examples.
However, I encountered an incorrect prediction. After doing some research and consulting with ChatGPT, I identified a few potential issues. One of the main problems is the small dataset used for evaluation. Additionally, the BERT model (bert-base-uncased) is pretrained on general text (such as Wikipedia and BooksCorpus), but it has not been fine-tuned for sentiment analysis on game reviews. This lack of fine-tuning leads to poor generalization, especially when dealing with informal gaming language. For instance, BERT may misclassify the word "punishing" as positive if it has seen it used in a positive context, such as in “a punishing workout.” Similarly, words like “challenging,” “difficult,” and “tough” might be seen as positive qualities rather than negative ones in this context.

If I had more time, I would address this issue by using a different pretrained model better suited for sentiment analysis, such as cardiffnlp/twitter-roberta-base-sentiment (which is trained on Twitter sentiment data) or nlptown/bert-base-multilingual-uncased-sentiment (which is fine-tuned on product reviews). This would likely improve the model’s ability to handle the specific language used in game reviews.


msharp9 (commented):
Nice job!  Your description is on point and I love that you did game reviews for Wukong.  Your tests and code are clean.

One note, your CI/CD pipeline failed since it couldn't find torch to run the tests.  You can add dependencies to the `uv` virtual environment used to run tests via the add command, e.g. `uv add numpy`.  For torch it's a little bit more tricky due to CUDA versioning, but details can be found here: https://docs.astral.sh/uv/guides/integration/pytorch/.  That said, `uv add torch` should add the CPU based version which would get your tests to pass.

PR #34, opened 2025-02-15
Nathan weber 4

Nathan-Weber208:
Sorry for the late submission, work has been hectic. 

Here ive used the standard evaluator that chat gpt recommends with some key differences.

I want to include a feature I heard about called "smart learning" where it chooses the tokens that have the lowest accuracy and emphasize that during training. in order for me to better train on the next loop i need to mark the portions that have the lowest accuracy.

My function returns those pairs. Now, i need to adjust the trainer to be able to take in the pairs and somehow emphasize them.
Of course this could lead to massive overfitting so i need to build it in a way it can work without so i don't have to rebuild. or maybe give it the ability to gradually increase the emphasis based on a hyperparameter? that's what i would do if i had more time.

### DIFF
--- a/dockerfile
+++ b/dockerfile
@@ -1,0 +1,20 @@
+FROM python:3.11
+
+# Set working directory inside the container
+WORKDIR /app
+
+# Install ZenML (before installing PyTorch)
+RUN pip install --no-cache-dir "zenml[server]"
+
+# Initialize ZenML
+RUN yes | zenml init
+
+# Expose the ZenML server port
+EXPOSE 8080
+
+# Copy required files (but NOT pytorch_packages)
+COPY run.py /app/
+COPY SW_EpisodeIV_VI.json /app/
+
+# Start ZenML and run the tokenizer pipeline
+CMD ["sh", "-c", "zenml login --local --ip-address 0.0.0.0 --port 8080 && pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu && python /app/run.py && tail -f /dev/null"]
--- a/run.py
+++ b/run.py
@@ -1,0 +1,380 @@
+import json
+from typing import Optional, List, Dict, Tuple
+import torch
+from torch import nn
+from torch.utils.data import Dataset, DataLoader
+from torch.nn.utils.rnn import pad_sequence
+from zenml import pipeline, step
+import torch.nn.functional as F  # for softmax
+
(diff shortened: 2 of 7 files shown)


dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# Expose the ZenML server port
+EXPOSE 8080
+
+# Copy required files (but NOT pytorch_packages)
+COPY run.py /app/
+COPY SW_EpisodeIV_VI.json /app/
+
+# Start ZenML and run the tokenizer pipeline
+CMD ["sh", "-c", "zenml login --local --ip-address 0.0.0.0 --port 8080 && pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu && python /app/run.py && tail -f /dev/null"]
```
msharp9, line 20: You'd probably save yourself some headache by installing torch with zenml on line 7 here.  That way it's installed during docking build, not docker run.

Docker cache's builds, so it'd still save time even with the .bat file.

msharp9 (commented):
This is great work, you really tried something ambitious too which is fun to see!  My only disappointment is that it was turned in late., but I understand.  Otherwise awesome job!

run.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+    Splits tokenized sequences into train, val, and test sets.
+    """
+    total = len(tokenized_sequences)
+    train_end = int(total * train_ratio)
+    val_end = train_end + int(total * val_ratio)
+
+    train_data = tokenized_sequences[:train_end]
+    val_data = tokenized_sequences[train_end:val_end]
+    test_data = tokenized_sequences[val_end:]
```
IMJONEZZ, line 91: Like we said in class, normally you want your test data to not be part of a statistical train_test_split, and even less so a non-statistical one, but no points off.

run.py
```diff
@@ -0,0 +1,403 @@
...
+    else:
+        padded_y = padded_y[:, :max_len]
+
+    return padded_x, padded_y
+
+
+# --------------------------------------------------
+# SIMPLE MODEL FOR NEXT-TOKEN PREDICTION
+# --------------------------------------------------
+class SimpleLanguageModel(nn.Module):
```
IMJONEZZ, line 169: Interesting, looks like you've looked at Chapter 9?

IMJONEZZ:
Solid work, thanks for the time and energy put in here!

PR #35, opened 2025-02-18
Homework assignment 5

tjkerby:
My plan is to train a LLM to predict cypher queries from text descriptions. There already exists a model that was trained to do this task: https://huggingface.co/neo4j/text2cypher-gemma-2-9b-it-finetuned-2024v1, but it is 9 billion parameters and I would like a smaller model that can easily run on a cpu and will have a smaller footprint. So I want to do knowledge distillation using that model as a teacher. After doing a bit of research I've decided that I will use https://huggingface.co/Qwen/Qwen2.5-Coder-0.5B-Instruct for my student model since it has strong performance for its model size and was trained specifically for code generation from questions which should help it learn the code-like language of cypher more quickly. For my dataset, I'll use the https://huggingface.co/datasets/neo4j/text2cypher-2024v1 containing just shy of 40,000 examples of a cypher query, its associated graph database schema, and a text version of the cypher query explaining what a user wants.

I will follow the guide in https://huggingface.co/docs/setfit/how_to/knowledge_distillation and page 182 from the class textbook for implementing this. 

For this, I will likely need to write a test to ensure that I handle dataloaders correctly, that I can load models in properly and get outputs using a forward pass, and that my loss function between student and teacher works as expected.


tjkerby:
After working on this for two days I've determined that this plan is not going to work out. The setfit package only works on sentence transformer models. I tried to get a custom solution working but my student and teacher models use different tokenizers, have different sizes of vocabulary, and have different architectures. Despite this, I still tried to make it work by doing prompt aligning, cross model projection, and gradient masking for token alignment, but I've been having complicated compile errors with torch dynamo that I haven't been able to overcome in an acceptable time period. 

So given this, I've decided to reconsider my approach. I could use google's 2B gemma model and still do knowledge distillation with that since I believe that they use the same tokenizer, but 2B is still larger than I was hoping for and for my use case I don't really love the terms of using their model. So my plan now is to use the Qwen2.5 coder instruct model and finetune it on the text2cypher dataset.

tjkerby:
Although my tests work just fine locally it seems that the way I have written tests fails on github since it doesn't have access to a GPU. I tried a quick workaround of skipping tests when a gpu is required but that produces a different error. So take that for what it is worth. I've attached an image showing it working on my end. I've already put so much time into this assignment that I'm not going to figure out how to get these tests working right now.

I probably should have realized this a while ago but it wasn't clear to me that these assignments were building off of each other and I was starting from scratch each time. So I have hopefully written my code a bit better to make it easier to build off of it. Regardless I wanted to switch datasets so I also switched tokenizers, among other things. 

My main.py loads the dataset from huggingface formats the data so that the text contains a brief system prompt, the graph database schema, and them the question that it is supposed to answer. I then load in the pretrained Qwen2.5-coder instruct model and tokenizer from huggingface and prepare it for finetuning using lora. I then put everything together, initialize the trainer, and begin finetuning. I've also attached an image of the training and validation losses as it trained.

Things certainly did not go the way that I planned and it is a good thing that I started working on it when I did. As far as my approach I tried to leverage huggingface's transformer and other available codebases to simplify things as much as possible. As I mentioned before, I ended up abandoning knowledge distillation and focusing on applying SFT using LoRA.

Since I started from scratch I wrote tests that hopefully covered my bases. I tested my data processing functions and made sure it worked on tokenizing it appropriately. Additionally I even have it do a step of finetuning and show that it has a train_loss and that the model parameters changed. I finally show that the model can be saved and that several of the appropriate files are successfully created.

If I had more time I would add more edge cases for my pytests, split my pytests up into different files, play around with changing my hyperparameters for LoRA (particularly r), and compare my models performance against the model I wanted to use knowledge distillation with.

image
image

msharp9 (commented):
Your tests are failing because no tests run. I do think the Unsloth GPU error in your test pipeline is an oversight on our part though. Overall great work. You'll be getting full points on the assignment.

As far as extra credit refactoring goes though, I don't see your evaluator, it appears you abandoned ZenML (which I guess is fine but would like to hear why since this is when it will come in most handy now that we are training), and I would have preferred to see your tests split out into different files.  Overall it's looking clean though and I'm just being a bit nitpicky here.

tjkerby:
@msharp9 I did end up abandoning ZenML, mostly since I don't understand the value it provides and currently feels more like a hassle than a benefit. I imagine that I'm wrong here and that if I understood it better I would change my mind. If it is an option I would love it, if you or chris could briefly cover this in class today.

IMJONEZZ:
Great work, I am really sorry that the assignments building off each other hasn't been clear until now. ZenML is a hassle for data scientists, but make MLEs' work much easier, we can address this in class today.

PR #36, opened 2025-02-18
Homework assignment 5

matthewwhite1:
Plan:

- I want to follow the same general structure of Chris' Unsloth code, but instead of using the school questions, I will use the Star Wars dialogue data
- The general idea is to have the output of the model be new Star Wars-like dialogue, trained from the actual dialogue data
- I will use Llama and Unsloth for training
- I will use SFT for training, which is common for LLMs - this will be easier than using something like RLHF, even though it provides less control
- I will use cross-entropy loss
- I will have to write tests to see if the ouptut dialogue is similar to Star Wars dialogue, although I will have to figure out how to do this in code rather than just looking at the output myself

(Edited this comment to add the bullet point about tests that I forgot to add)


matthewwhite1:
Things did not quite go how I initially planned. First, I ran into many problems with getting Unsloth to work because of Linux package dependencies (I use Windows). I figured some things out with that, but I wasn't able to fully use Unsloth. Then, after letting the model train for a few hours, I realized that it did not end up working how I wanted to, as the evaluator function showed me that the output was not what I wanted it to be (I ended up getting some "invalid parameter" messages). If I had more time, I would try to fix it and train the model again, but as of right now, it does not fully work.

Here's what the trainer attempts to do:
- Loads in Llama as the model (I requested and received access from Huggingface)
- Defines the training arguments
- Passes these arguments into an SFT trainer
- Trains the model and saves it to my computer

This time I had deepseek write my tests for me, and despite attempting some debugging, they do return errors.

I did refactor my code by moving things from the main file into their own files, so I would hope to get some extra credit for that.

It's a shame I couldn't train the Star Wars dialogue model correctly, but I think I will try to work on it more when it comes to my final project.

main.py
```diff
@@ -1,2 +1,29 @@
...
+
+
+@pipeline
+def assignment_5_pipeline(file_path: str):
+    """ZenML pipeline for text processing using BERT."""
+    data = load_data(file_path)
+    encoding = tokenize_text(data)
+    dataloader = create_dataloader(encoding)
+    # trained = SFT_train(dataloader)
+    # generated_text = evaluator()
```
msharp9, line 24: You'll want to uncomment these lines when you plan to actually train a model.

Also, you are tokenizing your dataset using BERT tokenizer, and then feeding the pre-tokenized into your trainer which is training a Llama model.  You then save the llama tokenizer with it, instead of the BERT tokenizer.  There are several issues with this, let me know if it isn't already clear now that I've pointed it out, happy to meet for some office hours.

msharp9 (commented):
Overall it's looking good.  Your trainer is straight forward, and looks good.  The tests for it though are a bit complicated.  I believe you are patching more than necessary and you have too many asserts which is a code smell indicated your SFT_train function is doing too much.  If you break it up into smaller functions it'd be easier to test. It appears your tests are failing and linter is also failing.  Also, a bit nitpicky, but in your plan, "I will have to figure out how" isn't a plan. :wink: 

Your refactor looks good overall, I would have liked to see a test for your tokenizer.py, it likely would have helped you catch your BERT/Llama Tokenizer issue earlier.

IMJONEZZ:
Keep in mind, the tokenizer must match the model, otherwise outputs will be crazy. Similarly, the embeddings must match the model and the tokenizer decoder must match the model too.

PR #37, opened 2025-02-21
Riley may assn5

rj-may:
My Plan of Action:

The instructions lack instruction 
"In your plan of action explain various approaches that you could take."  I am going to assume that this means the from step: 2 "Train a Model." I seriously considered jut building a classifier over 1 dimensional randomly sampled data.  That would by definition fit the question description and prove a point. 

If I am to train an LLM, I will do the following:

1. Redo my tokenizer from the last assignment using BERT on the star wars text data. 
2 Two options for  part 2
- A) copy the code from 5.2.1 of the book. This could be simpler, but working with code from scratch means that there are less nifty libraries to use. 
- B) Use unsloth do whatever Chris was saying if I get access to the notebook. This could be harder to work with something I haven't used before, but  a lot of stuff will be handled by unsloth I imagine
3) I pan to use ROGUE as a my loss function. It sounds more robust than anything I could make given my time availability>
4) Model architecture 
`Input data |>
tokenize |>
train and evaluate `

My test  will simply be to see if one training loop or step worked


rj-may:
## Explain your work 
#### How it went
I really didn't know where to start with training the model. I considered my plan of following the code of training from scratch, but I was reading the book and it says that training from scratch is "very expensive and time consuming endeavor." So I went down the route of trying to follow the provided code. My goal as to create some code that spoke and said stuff like star wars. I inputted the processed the star wars dataset , and I decided to train with that.  That was quite an adventure. I tried to train on Google Collab, and that did not work well. I ran out of free compute. I then had to do it on Kaggle. That took a long time (3+ hours). I should have tried to create something simpler, but I used the code given and just used a cross entropy loss function. I figured I was not trying to recall/ summarize, so I figured cross entropy loss would be better.  It eventually trained, and I was able to do some fun prompting. I got some fun prompts out. I  put those in a file called "star wars fun.txt"
 


 {"role" : "user", "content" : "Speak to me like Vader about the Force"},
'Your impertinence is... palpable.  The power of the dark side courses through me like a river of fire.  I am the embodiment of the Force, a master of its power, and a servant to none.  The dark side has consumed me, and I have become the very instrument of its will.  You dare to speak to me of the Force?  I am the one who has wielded its power to bring order to the galaxy, and I will not be lectured by a mere mortal.  Now, tell me... what do you know of the Force??"



#### If I had more time
I would want some measure of how much outputs are influenced by training. The loss went down (from 15 to 8), so obviously the star wars text had an influence.  
I think that overall, it  went great. I was very happy to just have some working code after so much time on it. 


#### Tests
For the test, I tried to use chat gpt to create some meaningful tests. 
All of the tests depend on  unsloth  and FastLanguageModel. This was difficult. The short story is that there is no GPU on my computer or on  this github to check tit. I couldn't figure out how to interact pytest with kaggle. I also didn't have time for anything like that. 


I spent the better part of whole day which I felt was more than enough for a one credit class, I have to call it quits  and move on to something else.

msharp9 (commented):
Fun results and great work. Don't worry about the failing Unsloth GPU error in your tests, that is a miss on our part.

Some notes, it looks like your finetuning was done relatively quickly.  Your loss stabilized way before you were done training.  I'd recommend looking into adding an early stopping callback.  Also to this comment: "The loss went down (from 15 to 8), so obviously the star wars text had an influence." Did it?  You started with Llama 8B, and I imagine it could produce similar results. A more robust evaluation methodology would be recommended.

I see you opted out of refactoring, but as a note, on notebooks.  I can't comment individual lines in a PR, so it can be confusing to give feedback. It's a best practice to still keep your core logic in python files you import, so I would have preferred to see a trainer.py that you imported into your notebook. Notebooks can import code from local files just as they can load data from local files, and save models to local files. If you did this, then your test_trainer.py would actually be testing your trainer.

Also, if you can refrain from closing your PR and let us do that I'd appreciate it. I almost missed this assignment.

IMJONEZZ:
Nice job, here's a note: your training took a long time because you were finetuning 8 billion parameters at 32 rank on a free GPU from 2015. If you want it to go faster, one simple solution is to make the numbers smaller, although the fact that you're able to measure the training time in hours is a huge privilege.

PR #39, opened 2025-02-22
Haozhe chen hw 5

chenhaozhe1:
Plans:

I planned to train and fine-tune a LLaMA-2 model, and I intended to add LoRA adapters to efficiently fine-tune the large model by adjusting fewer parameters. This reduces the memory and computational load, making it suitable for low-resource settings without compromising performance. Although I hadn’t decided on a specific dataset, I initially planned to use one of the built-in datasets for training. For the optimizer, I chose adamw_torch, known for its adaptive learning rate and weight decay capabilities. However, it can sometimes lead to overfitting if not properly tuned, particularly with smaller datasets or high-dimensional models. As for the loss function, I wasn't sure which to use in the LLM context. After conducting some research, I decided on causal language modeling (LM) loss, as it minimizes cross-entropy loss between the predicted and actual next token.

Works:

Since my device is a first-gen M1-chip MacBook from 2020, which has poor compatibility with many libraries (such as unsloth), I decided to do the training entirely in Google Colab using a T4 GPU. The implementation is in HW4.ipynb. I followed the plan, ultimately selecting the WikiText dataset for training. However, I encountered memory limitations on Google Colab due to the lack of a paid subscription, so I was only able to train on a small subset of the dataset.

My test function verifies the core functionality of the Trainer object from the main module, responsible for setting up and training the language model with specific configurations. The key aspects of the tests include:

Trainer Initialization: The first test (test_trainer_initialization_with_fake_data) ensures that the trainer is initialized correctly and has the required attributes, such as model, args, and datasets.
Dataset Validation: The second test (test_datasets_with_fake_data) checks that the datasets (train and validation) are correctly assigned to the trainer, ensuring proper integration for training.
Trainer Method Checks: The third and fourth tests (test_train_method_exists and test_train_method_runs_with_fake_data) verify that the train() method exists and is callable, ensuring it can execute without running a full training process.
Data Fitting Check: In test_fake_data_fits_in_trainer, I check that the trainer’s datasets are not empty and contain the expected number of examples, confirming the correct setup with mock data.
Training Arguments Validation: In test_training_arguments_with_fake_data, I ensure that the training arguments (such as batch size, learning rate, and number of epochs) match the expected values.
Trainable Parameters Check: The final test (test_model_trainable_parameters) ensures that the model has trainable parameters, validating that the fine-tuning setup is correct.


tests/test_trainer.py
```diff
@@ -0,0 +1,143 @@
+import pytest
+from main import get_trainer
```
msharp9, line 2: So this import statement is going to run your entire training pipeline.  It would be a good idea to encapsulate everything in a main() function, or behind a `if __name__ == '__main__':` statement

msharp9 (commented):
Hey, great work, I like your training pipeline.  Is there a reason you chose the WikiText dataset?  There's nothing wrong with this, but if you don't have a reason, we'd encourage you to use the starwars dataset to be consistent with the rest of the class.

I also like your plan, but it looks like you wrote it after doing the assignment.  That doesn't make it much of a plan now does it? :wink: 

Your tests failed because they couldn't find unsloth, remember to run `uv add unsloth`.  That said, you are safe this week because unsloth fails on imports when you don't have a GPU which is the case on Github. This was an oversight on our part. Lastly, there are some issues with your tests. I'm guess they pass, but they are sub optimal.  Be sure to mock when necessary.  You might want to also add a `!python -m pytest` command in your notebook to double check they run as a expected.  Similar to your `!python main.py` cell.

IMJONEZZ:
Your code looks incredibly familiar to me, which to me speaks to your experience working with LLMs. I'd love to see you push yourself a little bit more with the assignments, however I also understand it's a 1-credit class and you don't need to if you don't have time.

main.py
```diff
@@ -1,2 +1,90 @@
...
+    model,
+    r=8,  # LoRA rank (controls trainable parameter count)
+    lora_alpha=32,  # Scaling factor for LoRA
+    lora_dropout=0.05,  # Dropout rate to prevent overfitting
+    bias="none",  # No bias training
+    use_rslora=False,  # Keep standard LoRA
+)
+
+# Print trainable parameters
+model.print_trainable_parameters()
```
IMJONEZZ, line 30: How many did you get here?

tests/test_trainer.py
chenhaozhe1: Thanks for the feedback! I noticed that some packages worked well in Colab but aren’t functioning properly on my laptop, which might be causing the CI test failure. I’ll keep this in mind and work on ensuring the CI pipeline passes in the future.

main.py
chenhaozhe1: Here is my output:
trainable params: 19,988,480 || all params: 6,758,404,096 || trainable%: 0.2958

PR #40, opened 2025-02-22
Homework assignment 5

Nathan-Weber208:
Im planning on reorganizing my files, making seperate files for all the processes.

Most of all i want to make it so i can preprocess and tokenize the data, then train the model seperatly so i dont have to redo the first step. that means i will have to save it locally.

I want to use the bert model. mostly because im using their tokenizer so it will work well.


Nathan-Weber208:
took me forever but i got it working
the trainer is a adam w approach that spits out a loss record on ech epoch. it takes quite a while to run about 15 minutes. 
i would probably teach it to use the gpu if i had more time

dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+RUN yes | zenml init
+
+EXPOSE 8080
+
+COPY main.py pipeline.py data_loader.py tokenizer.py dataset.py model.py train.py /app/
+COPY SW_EpisodeIV_VI.json /app/
+
+
+CMD ["sh", "-c", "zenml login --local --ip-address 0.0.0.0 --port 8080 && pip install --no-cache-dir torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu && python /app/main.py && tail -f /dev/null"]
```
msharp9, line 15: You'll probably want to move your pip install torch up above, likely after line 5.  This way it's installed in the build instead of in the run command.

tests/test_main.py
```diff
@@ -1,11 +1,35 @@
...
+def test_split_data():
+    result = split_data(SAMPLE_TOKENS)
+    assert isinstance(result, dict)
+    assert all(key in result for key in ["train", "val", "test"])
+    assert sum(len(v) for v in result.values()) == len(SAMPLE_TOKENS)
+
+
+### 4️⃣ Model Training ###
+def test_train_model():
+    result = train_model(SAMPLE_SPLIT, batch_size=1, epochs=1)
```
msharp9, line 33: This test is going to run a full training loop.  Which is fine since you are just training BERT for a short time, but mocking might be a better solution.  I would also like to see more diversity in your tests to cover edge cases.

msharp9 (commented):
It seems like you focused on the refactoring aspect of this assignment. This seems to have come at a bit of a cost of the main assignment as we expect more from the planning and explain your work sections.  That said, I'm sure you made a wise decision.

Your repo looks much more put together.  I would like to see a lot more tests though.  There should be a test file for each of your main files and covering more cases.

IMJONEZZ:
> Im planning on reorganizing my files, making seperate files for all the processes.

Good idea

> Most of all i want to make it so i can preprocess and tokenize the data, then train the model seperatly so i dont have to redo the first step. that means i will have to save it locally.

You may not have to save permanently, there's a python package called tempfile that could help you here, and if that doesn't look like a win, you may want to look at the feather file format to keep things light.

> I want to use the bert model. mostly because im using their tokenizer so it will work well.

Solid idea, way to match tokenizers to models.

IMJONEZZ:
Your commit messages are nuts brother. You may want to look into testing your code before committing, that way you'll know whether the tests run locally, especially because it seems like the majority of the problem commits were formatting and linting.

PR #41, opened 2025-02-22
Elias thompson 5

eliasthompson304:
First off, I set to refractor my code so that my tokenizer, data loader, and evaluator were all put into their own files and imported into my final pipeline file. I experimented with kaggle a little bit which is why "use_model.py" exists. I still have my pipeline.py, however.

My trainer works by first obviously receiving training, validation, and testing data from my starwarsdataset. It then organizes each category into two subcategories: inputs ("Lines") and labels ("Characters"). I set up a neural network using torch.nn to turn data into vector embeddings as well as retain information for training using LSTM. 
Running with 10 epochs through the data I was able to compare the model's predicted labels or outputs with the actual labels given by the "train_dataloader." I followed a somewhat similar pattern for the validation data and printed out both the amounts of loss and accuracy each run through. After having saved the model path that is most accurate, I up returning it to plug into my evaluator. Now my evaluator script includes a predict function. Linking everything together in my "use_model.py" file, the user can enter in a line of dialogue themselves and get the model's prediction at which character would have likely said it. I'll attach an example below.

As I tested my code using pytest, I determined that 2 tests would be most helpful for me. When I first began seeing the results my model would give (right now I am doing a simple text classifier using the StarWarsDataset) they seemed too random to indicate any sort of learning. I wanted to ensure the model's progress would save and not just reset every time. I did this by running a test to check if the model path/file would actually save after each training run (right now I am just using a simple supervised training method. I tried to implement reinforcement learning but didn't quite get it working in time.) This test led me to edit my code to make sure the model_path was being saved each time.

My second test was to ensure the model would test consistently and not super randomly by using a different model file each time. I did this by simply comparing the model hash num before training and then after training. If they weren't the same I would know that the exact model pth/file wasn't consistent across training runs.


.DS_Store
msharp9, line 1: This file should be in a .gitignore, along with similar files.

.github/workflows/ci.yml
msharp9, line 1: You can't just delete this file, getting it to pass is the requirement for Question 3.

.github/workflows/pixi.toml
msharp9, line 1: why is your pixi.toml file in your github workflows folder?

USU-LLM-Class
msharp9, line 1: why did you add the repo as a submodule of itself?

best_model.pth
msharp9, line 1: models should be in the .gitignore file.

main.py
```diff
@@ -1,2 +1,2 @@
 def hello_world(name: str) -> str:
-    return f"Hello, {name}!"
+    return f"Hello and welcome, {name}!"
```
msharp9, line 2: having fun?

src/.idea/.gitignore
msharp9, line 1: This .idea directory should also be in the gitignore

src/_showyourwork.sqlite
msharp9, line 1: Why do you have a database?

src/train_model.py
```diff
@@ -0,0 +1,147 @@
...
+        model_path,
+    )
+
+    final_hash = get_model_hash("best_model.pth")
+
+    print(f"Model Hash Before Saving: {initial_hash}")
+    print(f"Model Hash After Saving:  {final_hash}")
+
+    # Check if hashes match
+    assert initial_hash == final_hash, "ERROR: Model file changed unexpectedly!"
```
msharp9, line 145: What the?

src/nn.py
```diff
@@ -0,0 +1,16 @@
+import torch.nn as nn
+
+
+class SimpleTextClassifier(nn.Module):
```
msharp9, line 4: We are really expecting you to use something better at this point in the class.

tests/model_checkpoint.pth
msharp9, line 1: There's a half dozen of these model files, remember to .gitignore models and datasets.

tests/test_trainer.py
```diff
@@ -0,0 +1,77 @@
...
+    first_hash = get_file_hash(model_path)
+
+    # train again
+    train_model_(process_data())
+
+    # get second hash
+    second_hash = get_file_hash(model_path)
+
+    # ensure the model file is identical acrosss runs
+    assert first_hash == second_hash, "Model file is inconsistent across runs"
```
msharp9, line 77: If the model file is the same before and after a run, that means the weights never changed. i.e. the model never trained.

msharp9 (commented):
Did you attempt to use an agent on this assignment?  If so, +1 for trying, I really like the gumption! Unfortunately, it seems you never created a plan, and you deleted the ci.yml file, meaning you didn't do half the assignment (ignoring the extra credit portion).

As far as refactoring and cleaning things up, there are many files that should be in .gitignore and many more that don't make sense why they were added or are just in the wrong location completely.  Note required for the assignment, but for the refactoring, I would have liked to see a test file for each major file not just trainer.py.  Most of your functions are also rather long and could have benefited from being shortened.

Also, since this is a course on LLMs, I recommend you start using one on future assignments.

PR #42, opened 2025-02-25
Homework assignment 6

eliasthompson304:


eliasthompson304:
I will use the AdamW optimizer, because of how it handles weight decay separately. I could use just Adam, but that wouldn't be the best since AdamW is basically the same except adaptive gradient updates don't mess with the weight decay. In the math for AdamW, in fact, it explicitly subtracts a scaled version of the weights at each step. 

I'm also going with cross-entropy loss for this assignment because I'll be fine-tuning an LLM. Predicting the next token will be important, and one can do that with cross-entropy loss. You could consider using mean squared error, but in this case because cross-entropy loss is used for classification in LLMs we'll go with that. Plus, cross entropy loss will allow for faster training since it is more sensitive to changes in predicted possibilities

I plan on sticking with the LLaMa 3.2 1B model's architecture (transformer based nn). Yes, fine-tuning will be necessary, however to avoid getting too complicated I hope to keep the model's architecture as is.

With these choices I plan on building my own dataset to either append to a general dataset or use by itself to accurately answer the questions. I will not train the model explicitly on the given questions.

msharp9:
Hey, I know you had limited time, but this is a pretty good start. There's only so much I can give you since you are missing tests and an explain your work section.  If you want to keep working on this, we can give you half credit for the half of the assignment you are missing.  Don't forget to add your model's responses to the questions if you do!

eliasthompson304:
Ok I will work on it some more today, thanks. I will add my model's responses to the questions once I get it right.

PR #43, opened 2025-02-26
Nathan weber 6

Nathan-Weber208:
My plans for this assignment are to go through the examples and try and integrate the fine tuning with my existing work.


Nathan-Weber208:
i decided to replace my tokenizer with the llama one just to match the model.

then i got it to run the questions. the answers to the questions are not "good" but they are what you would expect on something trained on starwars data then run on an existing llama model.

I also started using coda. so its using my gpu. if i had more time i would check the proccesses right next to each other to see which one actually saved me time.

msharp9:
> then i got it to run the questions. the answers to the questions are not "good" but they are what you would expect on something trained on starwars data then run on an existing llama model.

Just for clarification, it is assumed from the last assignment that you should know how to train a model on the starwars dataset. The object of this assignment is to get a model that can answer the questions. Essentially, just one step further. As discussed in class you will likely need to prepare a dataset to get decent results.

If you are looking for full points on this assignment, I would recommend by starting by creating a new plan.  The current plan shows zero thought and will get 0/3 points. There's still time.

msharp9:
Being lenient here, but I can't give you full credit for that plan of action, your tests are bare minimum, and you didn't include your model's responses to the questions in the homework assignment.

I'm assuming from your commit messages you might not realize that you can just run `ruff format` and `ruff check --fix` to fix 99.9% of linter issues. It's likely you are starting from an older branch so these commit messages are lingering from past assignments, in which case you might want to consider squashing your commits to clean up your history.

Nathan-Weber208:
I understand. It was a pretty rushed job to be honest.

I'll do better when I'm back in town next week.

Thanks for your time and work with us.

PR #45, opened 2025-02-27
Matthew White Assignment 6

matthewwhite1:
Plan:

- Since last week I used Llama and SFT, I should be able to follow a very similar setup this week, since the problem is pretty similar
- Instead of using the Star Wars dialogue data to train, I will write at least 10 questions that are similar but different to the provided questions, and these will be passed into the model to train
- I will use the same methods as last week (SFT, cross-entropy loss)
- I will write tests to ensure that the model is at least able to generate answers. I will have to decide for myself if these answers are correct


matthewwhite1:
As usual, this assignment did not go as planned. Here is a list of things a tried:

- I created a very small dataset with things that I hoped would help it answer the questions. After attempting to train on that, I definitely knew that this is not enough data to train the model.
- I looked into using distilabel to generate more data using my small dataset, but after running into error after error, I gave up on that.
- I looked for another question and answer dataset online, but since the one I found has quite different format to the Star Wars dialogue data, I could not get it to work.
- Throughout these processes, I ran into a lot of CUDA memory errors. My computer has what I thought to be a very good GPU, but I guess it might not be enough for LLM training. I think I might look into Google Colab for the next assignment to hopefully avoid these issues.

After all of these issues that took a while to work through, I decided to just call it a day. With more time, I think I would look more into using distilabel, but I just couldn't get it to work with the time I had.

msharp9:
You were actually very much on the right path.  You just needed a bigger dataset than what you created.  It would have helped if you included any of your distilabel work, you also forgot to include your model's responses to the 10 questions.  Lastly, don't forget to the format your code!

PR #46, opened 2025-02-27
HW assignment 6

rj-may:
I want to follow a very similar approach to the assignment I took last week. 
I want to take the code given in the notebook and give it some data that will help inform the prompts. 

I want to start with the wikipedia dataset. I thought that would help give some general knowledge to help answer the questions.  I want essentially feed it the wikipedia dataset, and see how it goes. If  that goes quickly, but poorly I might try and give it some  other.

I wanted to use the same defaults as last time such as a cross entropy loss function. 

For fun I gave the questions to copilot and it failed the president question. I don't expect to get that question right either, as I will be using the 2022 dataset.

I started messing with the code a little bit. I realized a few things. 

We are using  a different trainer
from trl import GRPOConfig, GRPOTrainer
instead of what I used  last time of 
from transformers import Trainer, TrainingArguments

Also, the wikipedia dataset  was super long. I am going to use the wikipedia simple to reduce it.

I got  to figure out some reasonable rewards because the GRPOTrainer requires rewards


rj-may:
Well, this plan has been a disaster. Trying to give the data in a good format has been difficult and having the right rewards

rj-may:
Things did not go as planned. The gist of my work on my last assignment was to use the star wars data to influence the prompts. I then decided to the same thing for this assignment. I thought to answer questions about knuckle popping, Talos, and sharks it needed more general information, so I would give it the wiki dataset. That turned out to be waaaay too big for me to work with so I went with the simple wiki dataset from hugging face. This solution with working the  wiki dataset had  a number of problems. One problem is it is from 2022, so it wouldn't get the president question right. Also, the text was way too long for the tokenizer setup (512 limit). I ended up just truncating it to the 512 tokens. That was probably not the wisest move, but it was a time saving move. Also, there was no way that it was going to answer some of the other logic questions correctly, so I decided to give it a jeopardy dataset and a riddles dataset. 

https://www.kaggle.com/datasets/aravindram11/jeopardy-dataset-updated
https://www.kaggle.com/datasets/prajwaldongre/riddles-a-synthetic-riddle-dataset-for-nlp

These datasets may have helped with some knowledge and logic, but it probably needed more time to train, and I don't have  more time to give to this class. This class is interesting, but I don't have the time to dedicate to complete these all the way. 
Anyways, the results I got using the combination of three datasets were ok in terms of accuracy. The sharks one tells me that sharks don't swim in pools, but also to call a professional snake hunter.  The sphinx question was interesting to read. There may have been too many riddles surrounding coins, so it got really excited to talk about coins as a possible answer. 
Overall, the model does understand the wood chuck, Talos,  and pound questions. 

I would live to give it the full wikipedia dataset for more knowledge, but I don't if that would have helped. 
This part of the assignment was interesting, and it took me basically all day on Thursday. 



I never really got my test trainer working. I haven't quite figured out pytests and doing those on google collab. I  am also tired of installing all these packages on my computer such as vllm just to run pytest. My py tests are mostly the work of chat gpt, I ran out of time for this.  I dedicated Thursday to this project, and that wasn't enough time as I had lots of bugs with the wiki data. So it is eating into my other classes here on Friday, so this is where my work ends. 



Overall, I like my answers.txt file. It told me  it was using the data that I gave it. The data I gave it didn't make the answers come out exactly, but it was fun nonetheless. 

Maybe I should have just trained on questions super similar to the 10, but that would have felt like cheating.

msharp9 (commented):
I enjoyed reading your LLMs responses to the questions! I think you are beating yourself up a bit, because this was just a difficult assignment. You were able to get a few of the questions correct and should be proud! It might not seem like much, but very very well done.

PR #48, opened 2025-02-28
homework assignment 6 - Plan of Action

jashonnew:
- In terms of approaching finetuning our model, I think there are two main variables we are able to adjust here in order to finetune the model. We can choose a method for finetuning, and we can choose a dataset. We could have an LLM create the data needed in order to answer the questions correctly. There is most likely a relevant Wikipedia article for each of these questions, or something Wikipedia-esque could be created in order to give the model more context for answering these questions. We could also use some sort of distillation fine-tuning where we take a model that can correctly answer the questions and have combat our model to help it  understand the logical deficiencies in answering.

- I will use Adam as my optimizer, or something else off the shelf.

**Which loss metric will you track?** 

-I will track cross-entropy or something else off the shelf.

**Which model architecture will you use? Talk about their potential pros and cons.** 

-We are supposed to use Llama 1b so that is what I will attempt to use.

**What alternatives did you consider to these decisions?**

- I considered SGD and KL Divergence for the first two, I didn't consider any other models as it was specified for the assignment.


jashonnew:
I just used a manual training method on the Llama 1b from huggingface. It doesn't work fully, and I ran out of time in my 3 hours for homework before I was able to debug. I was also not able to verify that my tests are working. If I had more time I would certainly get my jupyter notebook (the notebook instance i ran was via kaggle for GPUs). I was also not able to get chris' notebook running and i wanted to understand that process as well. Only so much time in a week unfortunately.

msharp9:
This is a good way to maximize points while minimizing work, I like your style.  Your plan is good, and your explain your work section is honest enough.  A couple comments, we asked for `meta-llama/Llama-3.2-1B`, not `TinyLlama/TinyLlama-1.1B-step-50K`, using a pre-trained model instead of the foundation model, could be a good shortcut so not sure if that was your intention.  It just seems like you weren't able to load in the TinyLlama.  I would have been curious to see your results, since you decided to feed it the actual questions which we instructed you not to.

Also, next time remember to format your code!

PR #49, opened 2025-02-28
Tom kerby 6

tjkerby:
My plan is to finetune the Llama3.2 1B model on a custom dataset. To create the dataset I will create a custom dataset where I create variations of the questions given. Here is an example question:
"If a movie lasts 2 hours with 20 people watching it, how long will it take to watch the movie if 200 people watch the same movie?" with the answer being "2 hours". Hopefully this highly curated dataset will help it answer the questions without too much trouble. I will follow Chris's notebook in https://www.kaggle.com/code/chrisb304/chris-llama-grpo fairly closely for structuring my code  and working with vllm for finetuning.


tjkerby:
[answers.txt](https://github.com/user-attachments/files/19036949/answers.txt)
This was a hard assignment for me. I was able to finetune my model using my custom datapoints but as you can see from the answers it gives it didn't perform as well as I would like. 

I originally tried using GRPO as was used in the kaggle notebook but this seemed to result in catastrophic forgetting where the model output gibberish. After switching to SFT it did a bit better but not much. 

I spent probably at least 2 hours trying to get distilabel working to generate additional synthetic data, but I couldn't get it figured out. You can look at my branch to see my attempt. If you know where I went wrong I would love to learn how to do it properly.

I didn't write any additional tests for this as I don't really feel like I got the hang of it and I've already invested far more time into this than I anticipated needing (6+ hours total).

If I had more time I would really have liked to figure out distilabel and get a synthetic dataset with a few hundred to a few thousand observations. Additionally, I would have spent more time organizing my code to split up my tests, write a few more tests, and get zenml working again with my code.

msharp9:
Hey this was good work. It was a difficult assignment so I wouldn't beat yourself up over not getting any of the answers right. You were on the right track but didn't know it.  Good job on realizing you needed to switch from GRPO to SFT. That was a smart move. While reasoning could help a model answer some of these questions, a 1B model just doesn't have enough context to properly reason about the questions we gave.

PR #51, opened 2025-03-10
Assignment 7 Vector DB PR - Plan of Action

jashonnew:
- Build a vector database

1.  I would like to use Pinecone and cosine similarity. Pinecone because the implementation seems a bit more straightforward for someone without the architecture experience like myself, and cosine similarity for its magnitude independent similarity assessment.
2. As alternatives I considered Milvus for pinecone, and Euclidean distance/ dot product for similarity measures. For the reasons llisted above, I didn't move forward with those options.

- Query it to make sure that it is working 
- Make sure that my instance of trained LLM 1b can access the database and is attempting to access it before answering questions
- Ensure that the content of the vector database makes sense in relation to what the LLM is able to pull
- Adjust as needed from there to optimize the amount of questions answerable by the LLM
- Write tests to ensure any problems that I faced work into the future, try to predict what issues the system will have at scale, write tests for these issues specifically. 
- Along the way I will need to frequently reference online resources as well as chapters 3/6 of the LLMs in production book

Full Pros and cons for cosine similarity in a vector DB:
**_Pros:_**
**Scale-invariant**
**Useful for text embeddings:** Frequently used in NLP applications where TF-IDF, Word2Vec, or transformer embeddings represent documents.
**Works well with normalized vectors:** Many embedding models, such as Sentence Transformers, return normalized vectors where cosine similarity is a natural choice.
**_Cons:_**
**Ignores magnitude:** If magnitude carries meaning (e.g., word frequency in TF-IDF), cosine similarity may lose valuable information.
**Computationally more expensive than dot product:** Due to normalization, extra operations (vector norm calculations) are required.
**May not be the best for dense retrieval in some high-dimensional spaces:** Sometimes, using dot product directly can be more efficient.


jashonnew:
I know you guys dont prefer ipynb, but i successfully implemented a pinecone cosine similarity vector db and got 8.5 of the questions answered correctly by llama 8b. Not sure how you want to grade this necessarily, but it seems like at least some of the extra credit points should come my way. Fun assignment. I implemented it with the live model, so in this sense tests didn't make sense. But the test of having it improve the answering of an LLM worked great!

msharp9:
Great work! It was fun reading through this PR.  I essentially gave you full points at 50% off for what you turned in late.

As a reminder, we covered in class some strategies to write tests for those who prefer working in jupyter notebooks, it might be worth reviewing that.

PR #52, opened 2025-03-11
Homework assignment 7

rj-may:
My plan for this commit.

For creating a vector database, I have  a number of thoughts.

The big goal of this assignment is to create vector embedding. The example code provided works on the wikipedia dataset, and that might be rather large for the context of this assignment. I think I will work with something. I am going to mess with one of my final project ideas of using all of yoda's speech from all 6 star wars movies.  That should be a nice small.

No, I have to choose something way to vectorize things. I could use the example in listing 6.7 of a pinecone database, but that I don't want to deal with the API, and it seems excessive for my needs. I am instead going to look at the Chroma example given in the assignment description. If that doesn't work, I might go try something like FAISS from Facebook. We have been using a lot of meta stuff with llama, so that could be good.

When I get to querying, I hope to use cosine similarity. This feels intuitive for a vectorizer. Alternatively, I could use a nearest neighbor approach.

Hopefully, this should be fairly simple.


rj-may:
This assignment was  lot less heavy with the code then other assignments. That was nice. But I struggled to test my code without a full running implementation. I relied on chat gpt for the tests.  The pytest mocks were tricky to work with. I wanted to test to see some generic things like what happens if I don't input anything in the vector database what does it return.  

Overall, it appeared to work pretty well.  With the prompt, 
"Tell me about the dangerous Sith"

[' Destroy the Sith, we must.', 'The very Republic is threatened, if involved the Sith are.', ' To fight this Lord Sidious, strong enough, you are not.']

After working on this project, I am not sure if the yoda plan will make for a good final project. There was only 100 yoda lines in the dataset I  was working with.

PR #55, opened 2025-03-12
Matt Burns Assignment 7

m6urns:


m6urns:
#### Development Plan

For this assignment, I plan to implement a vector database component that integrates with my existing LLaMA fine-tuning pipeline. After reviewing the codebase, I've decided to use ChromaDB for local development. This lets me build out the system without worrying about usage costs during development. I might consider Pinecone later as a secondary option if the chromaDB option goes well. 

I'll create new ZenML steps for embedding generation, database operations, and retrieval functions that work alongside my existing pipeline structure from Assignment 6. I am hoping with this method to be able to successfully address the questions from last weeks assignment. Starting with ChromaDB locally gives me independence during development, and I can adapt as needed make needed additions to my current codebase.

m6urns:
For this assignment, I implemented a vector database integration with my Llama 3.2 fine-tuning pipeline using ChromaDB as the backend. I created a well-structured abstraction layer with VectorDatabase as the base class and ChromaVectorDB as the implementation, making it easy to swap out for other vector DB solutions if needed. I did not end up creating an implementation with Pinecode. The embedding generation is handled through SentenceTransformers with the "all-MiniLM-L12-v2" model by default. The system allows for adding texts with metadata, searching for similar content, and performing RAG during model inference. This solution would still require more work to successfully answer the test questions. I was able to put together an effective method to clone and run directly from my repo on a Colab instance which made the processes of training and testing significantly easier compared to the previous assignment.

PR #56, opened 2025-03-13
initial commit

Nathan-Weber208:
My plan of action is to first get the llama Model from last assignment running (i struggled on that) 

I will run the same questions through it.

then i will add a vecorDB (specifically the Wikipedia one from the examples) and run the questions through again.

then i will have 2 sets of questions i can compare how well they did against each other

If the agent really does help it, then we should see drastically more accurate answers to the questions.


Nathan-Weber208:
Im really struggling to understand how to use zenml

I ended up just scrapping it all and keeping the part where i actually build a super simple agent.

I need to connect with one of my classmates about zenml and go over how exactly it works.

PR #57, opened 2025-03-13
Homeowkr Assignment 7

eliasthompson304:


eliasthompson304:
Plan of action:

How to complete requirement for question 2: create a vector database.

I will create my vector database using Pinecone. According to my research, Pinecone is fully managed in the cloud (no need to worry about infrastructure or setting up my own server). With Pinecone I won't need to manually store the index for my vector database - it will already be hosted for me. In addition, Pinecone automatically scales to handle large amounts of vectors without the need to upgrade hardware (I may not have to worry about this problem anyway, however). Pinecone also allows for real-time "upserts" without slowing down queries. I can easily access Pinecone via API, from anywhere.
I'll be attempting to improve my dataset from the last assignment and then use this dataset to create my vector db.  

How to complete requirements for question 3: write tests and format your code.

I will look to test 5 things to ensure my Pinecone vector db works as expected:
1. Index Creation - test to see if an index is successfully created (if not nothing else will work)
2. Data Insertion or what could also be called "upsertion" 
3. Querying 
4. Edge Cases for empty queries or non-existent IDs

At the end of this assignment - if i have time - I will also attempt to use what I 've learned to answer more of the 10 questions from last assignment.

eliasthompson304:
After spending some time reading pinecone documentation I was fairly easily able to create a vector db using it's easy start guide. Then the real problem became reformatting my dataset so that is worked with pinecone's requirements. After refactoring my dataset with ids, types, chunk_text fields, I was able to get the vector db stats printed out as well as a query going where the user enters in a query and gets the closest matches from the dataset. Here is a screenshot of it in action: 
Screenshot 2025-03-14 at 9 42 45 PM

As far as tests, it was still a learning curve for me to write tests that run correctly. I had an idea of what kinds of tests I wanted to run in my plan, but I ended up getting as far as two tests: checking if the pinecone logic used actually creates an index and whether upsert and query works correctly.

PR #58, opened 2025-03-14
HaozheChen Assignment 7

chenhaozhe1:
Plan of Action: 

For this assignment, I’ll be using the Pinecone API to build a vector database with the WikiText and BookCorpus datasets (starting with the first 1000 examples, though I might try to include more depending on the time I have before the deadline). Once I’ve got the database set up, I’ll use it to improve the training of the model from my last homework assignment and see if I can improve how it answers the questions.

I’m going with Pinecone because it’s really good at handling large-scale vector searches, which is perfect for things like semantic search and improving AI training. Plus, unlike other APIs like FAISS (after some research), Pinecone automatically scales based on usage, so I won’t have to tweak anything manually. 

For the similarity metric, I’m planning to use cosine similarity. The great thing about cosine similarity is that, unlike Euclidean distance, it’s not driven by the magnitude of the vectors. Instead, it focuses on the angle between them, which is helpful in a language model setting where word embeddings might have different magnitudes but still have similar meanings.


chenhaozhe1:
After working on this, I couldn’t get the vector database to work with Llama-3.2-1b because it requires an output dimension of 128256 for Pinecone. I tried adding a hidden layer to map the output to something more reasonable, like 1024, but ran into some technical issues. I think I could fix them in the future, but since there was a time crunch for this assignment, I’m just submitting the first part—setting up a vector database and querying it with the same questions from the last assignment.

You can find the full implementation, including the pytest tests (all passed), in hw_7.ipynb. For embedding, I used "squad-text-embedding-ada-002", which is based on the Stanford Question Answering Dataset (SQuAD). Since this dataset has tons of factual Q&A pairs, I figured it would be a good choice for learning factual knowledge. As for the similarity metric, cosine similarity wasn’t working well—it gave really low query scores (usually below 0.1), so I switched to dot product, which made a big difference.

For testing, I wrote a script that checks if the Pinecone-based vector database works properly. It uses mocking techniques to fake index creation, data insertion, and querying, so we don’t need real API calls. The test makes sure the index starts empty, supports batch data insertion (without actually storing anything), and returns expected query results using a pretrained Sentence Transformer model. This way, we can validate the logic before deploying with actual Pinecone services.

IMJONEZZ:
128256 isn’t the output dimension of Llama, it’s the context window. You may want to look at the d_model parameter to figure out what is going on and how you can change it instead of trying to add hidden layers (which will change model output).

Other than that, amazing work, very happy with the vectorDB.

PR #59, opened 2025-03-14
Matthew White Assignment 7

matthewwhite1:
Here is my plan for this assignment:
- Do some more research on what a vector database is
- Look up some python examples of a vector database
- Think about how I want it to be implemented into my current architecture
- Have a discussion with ChatGPT about how to accomplish this
- Create a vector database file that uses the tokenizer to create vectorized embeddings of the data
- Query the database


matthewwhite1:
As always, things didn't go as planned. ChatGPT helped me with the coding part, and when I tried to run the code, it would always freeze at my get_embeddings function and freak out my entire computer. I started this a bit too late to put more time into debugging, so I am just going to leave it as is. My vectorDB code creates vector embeddings of text, calculates the distances between the vectors, and then prints the embeddings and distances. As always, ChatGPT wrote the test file for this assignment.

evaluator.py
```diff
@@ -0,0 +1,60 @@
...
+    AutoTokenizer,
+)
+
+
+def evaluator():
+    """Generates text using the trained LLaMA model with proper error handling and tokenization fixes."""
+
+    # Load the trained LLaMA model and tokenizer
+    model_path = "assn_6_llama"
+    tokenizer = AutoTokenizer.from_pretrained(model_path)
```
IMJONEZZ, line 14: This may be why your embedding function is freaking out, I’m not seeing model_path being defined anywhere to actually do the embeddings, including in main.

tokenizer.py
```diff
@@ -0,0 +1,15 @@
...
+from transformers import AutoTokenizer
+
+
+@step
+def tokenize_text(data: list):
+    """
+    Tokenizes text using Llama's tokenizer and returns tokenized tensors.
+    Assumes each item in data is a dict with a "Line" key.
+    """
+    tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-1B")
```
IMJONEZZ, line 11: Seems like you’re loading this several places throughout the code

trainer.py
```diff
@@ -0,0 +1,61 @@
...
+
+@step
+def SFT_train(train_dataset):
+    torch.cuda.empty_cache()
+    torch.cuda.ipc_collect()
+
+    tokenizer = AutoTokenizer.from_pretrained("meta-llama/Llama-3.2-1B")
+
+    model = LlamaForCausalLM.from_pretrained(
+        "meta-llama/Llama-3.2-1B",
```
IMJONEZZ, line 22: Also loading the model a couple of times

IMJONEZZ:
All in all, great work. Sorry it didn’t all go as planned, but hopefully this task was a nice change of pace.

PR #61, opened 2025-03-19
Assignment 8 POA

jashonnew:
After reading chapter 8, my plan of attack is as follows:

- Familiarize myself mega with the hugging face hosting capabilities and platform. Try and see how models can be accessed, how my own model could be hosted, how I can access that model via an API, how I can use an API to reference my app and host something locally/semi-locally. 

- I would like to try out either the streamlit or the gradio examples from the book. I know that the code is hosted on Chris' github and that would be an accessible place to start. My only hurdle would be to get a working model api, and get my model potentially into that api location. I think gradio may handle many edge cases for me, but the streamlit example is already provided front to back. I think that hf spaces has some gradio integration so that may be an easy place to start (especially if code is provided).

- Read some front to back examples of LLM applications that others have made/hosted. It seems like commonly some of the necessary nitty gritty of the architecture is left out of walkthrough examples, so I will have to collaborate with some folks who are more comfortable with hosting in python. 

- A big piece of what is missing with my plan, is what my app/model should do relative to models out there, like what is the task I am trying to do or problem I am attempting to solve. This is still a bit unclear to me. I need to get past the anxiety of being able to do the thing before I can open myself up to creatively deciding what I would like my app to do. 

- Lastly I need to understand curl requests and make sure that my application is formatted in such a way that it accepts these requests.

### DIFF
--- a/Application.py
+++ b/Application.py
@@ -1,0 +1,32 @@
+from flask import Flask, request, jsonify
+from gradio_client import Client
+
+app = Flask(__name__)
+
+# Initialize Hugging Face Gradio client
+client = Client(
+    "Jashonnew/deepseek-ai-DeepSeek-R1",
+    hf_token="your_hf_token_here",
+)
+
+
+@app.route("/chat", methods=["POST"])
+def chat():
+    data = request.json
+    user_input = data.get("message", "")
+
+    if not user_input:
+        return jsonify({"error": "No message provided"}), 400
+
+    # Send message to Hugging Face model
+    try:
+        result = client.predict(user_input, api_name="/apply")
+        print(f"API Response: {result}")
+        return jsonify({"response": result})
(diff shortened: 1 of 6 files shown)


jashonnew:
Image below with curl requests. I was able to get it to successfully hit the API, and I was able to get answers out of my hosted deepseek model on hf spaces, but I could not get anything but blank returns via the API. I need someone else to look at my code at this point and have spent about 7.5 hours working on this assignment. 

The gradio files are what I was using to run my huggingface space, i copied them into the repo so you could see how i was running it. I tried making things more complicated in order to get more API endpoints that would hopefully work, but was unsuccessful. All code was formatted with ruff. 'Application.py' was me connecting to the API endpoint before sending the curl requests. 

image

PR #62, opened 2025-03-20
Elias thompson 8

eliasthompson304:


eliasthompson304:
Plan:

Before I go through the effort to serve my model I want to fine tune it a bit more. I am going to use Gutenberg or Short stories dataset that will correlate with an idea I've had for a final project (an interactive story teller). After I finish fine tuning the model - whether it works as intended or not - I will use hugging face to serve my model. I will create a new model repository and then deploy it using FastAPI. Then use curl to send requests

Pros - serving my model on hugging face seems to be a simple process.
Cons - I can't think of any atm,

eliasthompson304:
Screenshot 2025-03-22 at 12 01 07 AM

So I followed my plan and attempted to fine tune the llama model over again.. too bad it didn't turn out very well haha. 
I did however get the local server up and running and was able to use curl to send requests. I have a screenshot attached below. I would improve this pr by going through my data prep again as well as my training loop to ensure everything is set up for optimal results.

eliasthompson304:
I meant to attach this screenshot as well last night:

PR #63, opened 2025-03-21
Tom kerby 7

tjkerby:
My plan is to follow the example in https://github.com/IMJONEZZ/LLMs-in-Production/blob/main/chapters/chapter_6/listing_6.7.pinecone.py fairly closely. I like the idea of using a subset of wikipedia as my dataset as it seems like a nice general purpose dataset for augmenting an llm and giving it the ability to potentially cite sources. Since I don't want to spend money on OpenAI credits I will use a huggingface sentence transformers model to create my embeddings and pinecone to store and query them. From a quick search it seems like all-MiniLM-L6-v2 is a good general purpose model for embedding text so I will plan on using that one. Since wikipedia articles can be fairly long and my model is meant for encoding paragraphs not entire documents I will follow the code example and use a text splitter to split the articles up into chunks and save off the appropriate metadata to allow a user to find the rest of the article if desired.

For testing I will test that the embedding model works as expected and I will additionally test my wikipedia data digestion process.


tjkerby:
Things went fairly smoothly. I was able to use pinecone relatively easy and was able to load and use a subset of the wikipedia dataset with the huggingface sentence transformers model for embedding. Chunking wasn't much of an issue and I just needed to be careful to chunk them up into small enough chunks to fit into the sentence transformers context window (256 tokens) to avoid being truncated.

I wrote my tests largely using help from perplexity ai. In it, I do some basic tests to see that my WikiDataIngestion is properly inialized, that splitting returns the proper output, that batch encoding works properly, that uploading to pinecone works, and finally that you can properly query from it.

If I had additional time I would focus on trying to do the extra credit and design the vector database to help answer the questions. I might also experiment with semantic text splitters rather than doing it by a token count.

PR #65, opened 2025-03-21
Matt Burns Assignment 8

m6urns:


m6urns:
I plan to implement a FastAPI based server ontop of my existing query_model.py script. I have previously utilized FastAPI and it's familiar. I plan to implement an API, a simple client to perform queries, and to document usage with CURL. I will also implement error handling and parameter validation.

m6urns:
This went very well if I had expected it to go this quickly I would have tried harder to get it in on time. I was able to create a simple server that can be queried via curl. The implementation uses FastAPI for automatic documentation. I was able to serve and query a locally trained model. The server includes request validation. 
Screenshot from 2025-03-25 19-26-11

PR #66, opened 2025-03-21
Tom kerby 8

tjkerby:
My plan is to try to use ollama to serve my model. This means that I need to be able to make a Modelfile for my finetuned model. Doing some quick research on the subject let me know that I need my file to be in a GGUF format in order for it to work with ollama. So I need to convert my finetuned model to be in GGUF format and a quick search leads me to look into llama.cpp to do that. Assuming that the process goes well without hiccups I will then create the model file upload it and be good to use it hosted on ollama. If I run into issues doing this I will get off my high horse and use huggingface inference endpoint. I'm avoiding it since they want a credit card on file to use it and I try pretty hard to avoid that when I can.


tjkerby:
Here is the curl output:
image
image
image
image
image
image
image
image

Sorry that this is streamed and as such is spread out so much. But it does respond properly if you take the time to look through it. I'd fix it before I submit but given the hour I'm pooped and calling it a night with this.

tjkerby:
Things did not go how I planned. I got hung up for a long time trying to figure out how to properly save my model as a .gguf file so it would work with ollama. Running model.save_pretrained_gguf("model", tokenizer) kept running into issues as it was trying to install llama.cpp but it couldn't seem to build it properly on its own. When I tried to install it manually by running cmake -B build it would complain and throw errors. I went down a long rabbit hole and finally found out that the issue is that although I had properly reinstalled cmake due to caching issues it wasn't seeing the new version and was still breaking on the old version. I finally got this fixed by running hash -r and then it worked beautifully. Once this was done I made a Modelfile pointing to my .gguf file and then I could run my curl command. 

If I had more time I would work on making the output prettier for dealing with streaming. Additionally, my model expects <|im_start|> and <|im_end|> tokens and they are a pain to input when you are writing a query. I would add a function to insert these in when you query the model.

PR #67, opened 2025-03-21
homework assignment 8

rj-may:
Making a plan just doesn't work when you have no clue what you are doing. So my plan comes after experimenting and failing some. 

**The big goal of mine is to take my model from assignment 5 and find a way to deploy it**

First, I wanted to use Ollama, but I read something that it wouldn't take my custom model. So, that won't work. 
I am now attempting something on Collab. I am trying to use VLLM. 

I am waiting on access (as I write this) for access on hugging face to use meta-llama/Llama-3.1-8B-Instruct. I hope that  happens soon. I didn't know I would need to do that, so that would be hard to plan for. Hopefully, this works. 

I will be using the documentation from these two + ChatGPT to figure out the rest of this. 
[vllm](https://docs.vllm.ai/en/stable/features/lora.html)

[meta llama](https://huggingface.co/meta-llama/Llama-3.1-8B-Instruct?local-app=vllm)


rj-may:
So because I trained with 
load_in_4bit= True

apparently I cannot host it. This is weird and frustrating.

rj-may:
I tried a lot of different things on this assignment. None of them really worked. I think I exhausted plans a-m, and it time to retire for the night. 

the file assn8.ipynb represent most of my initial efforts. I tried a bunch of different things and debugging. The setup here is to use VLLM to host it locally. I have the **requests set up**, it just won't load. The only thing I could figure out is when I trained my model, I had  load_in_4bit= True which messed up everything.  Apparently vllm doesn't support that. I  am so confused, but I have all the framework there.  I went through this whole rigamarole of getting an access token and requesting access to llama model. 

I also tried retraining the model without load_in_4bit, but collab must have changed some thing with numpy and I oculdnt figure it out. 
ValueError: numpy.dtype size changed, may indicate binary incompatibility. Expected 96 from C header, got 88 from PyObject


I also tried using fast api but I ran into the same numpy error this is seen Colab_LoRA_Unsloth_API(1).ipynb code.  


i don't know what I would do if I have more time. I guess I could try and host something else, but I don't know what.  I learned a lot but I was frustrated by more.

PR #68, opened 2025-03-22
Plan

Nathan-Weber208:
1. Use last weeks code to have a working LLM model
2. Rewrite it to use a user input instead of a fixed strings
3. Serve it up using flask
4. Vibe code a front end for the model
5. Make it into a docker container
6. Deploy it locally


Nathan-Weber208:
I ended up not using flask but fastAPI. that's what the example recommended and i ended up learning a bit using it.

I wrote a test that actually spun up the app and ran it because that is what i was most focused on this assignment. I know that the model works i want to be sure the server works.

If i had more time i would expose my GPU to the docker image. then i could really get some speed out of it.
This method takes a good 90 seconds to get a response if it doesn't find a cheat in the vector DB.

PR #69, opened 2025-03-22
Matthew white assignment 8

matthewwhite1:
Here is my plan of action:

- Figure out what serving a model really means
- Realize that since my model didn't train correctly a few weeks ago, I may not be able to successfully do this assignment without re-training the model, which I do not have time to do right now
- Look into if it's possible to serve a different model
- Have ChatGPT help me write some code anyway of what I could do to hypothetically serve my model
- Attempt to send requests to the server

That's all I got.


matthewwhite1:
Unfortunately, I really did not leave myself enough time to do this assignment - I am realizing it is a bit too much mentally for me right now. Something I want to do this week is rerun my Llama model based on the Star Wars dataset and see if I can get it to work. My grade is still ok in this class so I am willing to take a fall of a few points.

PR #70, opened 2025-03-22
Erick graham 8

Chazzwazzler:
I plan to use Ollama to serve my model. I thought about using Huggingface's inference endpoints, but it requires my card information so I would rather use something else, though it may require more work.

I will have to create a Modelfile before I can create a model using Ollama, so I took a look at the Modelfile documentation. It appears most of the information the Modelfile needs is just customization and parameters (stuff I will just have to adjust as I work on this), aside from the FROM field. I can use my own model for that by converting it to a GGUF file. That process appears to be surprisingly easy, I can just use llama.cpp through docker to convert a model to the file format I need. After doing so, serving the model just requires a few Ollama commands:
`ollama create mymodel -f ./Modelfile` (create a model)
`./ollama serve` (start the server)
`./ollama run mymodel` (run the model in the server)

I think that is all I will need to do to serve it. The Ollama documentation includes how to send curl requests to the model as well, which takes care of question 3, after I use ruff to format the code.

### DIFF
--- a/Modelfile
+++ b/Modelfile
@@ -1,0 +1,1 @@
+FROM Llama-3.2-1B.F16.gguf
--- a/main.py
+++ b/main.py
@@ -1,3 +1,3 @@
 def hello_world(name: str) -> str:
     return f"Hello, {name}!"
- 
+  
--- a/main.py
+++ b/main.py
@@ -1,2 +1,3 @@
 def hello_world(name: str) -> str:
     return f"Hello, {name}!"
+ 


Chazzwazzler:
image
image
image

Chazzwazzler:
Things generally went well, and the server is working. I had a bit of difficult converting the model into the file type I needed, but I eventually figured it out. After I got that working, I was able to get it working just fine with the commands in the Ollama documentation.

PR #71, opened 2025-03-26
Elias thompson 7

eliasthompson304:
removed api key


msharp9:
Just as a note, even removed it was still exposed, and still in the git history.  Just make sure you rotate the key, (e.g. go to Pinecone and delete it there.)

PR #72, opened 2025-03-27
Initial POA PR commit - assgn9

jashonnew:
After reading in both chapter 7 as well as the blog posts provided, my plan for enforcing constrained html output from the LLM is to use an the guidance package so that the output is forced into the html format required. If there is extra time I would also like to use guidance to try and give my model access to a calculator. The model that I am using is still a Deepseek through Huggingface Spaces. I may try switching to outlines if I get stuck but the guidance documentation seems to be a bit more clear. I will also reference the documents Chris shared with us if I need to get more ideas or start from scratch.

### DIFF
--- a/HF DS Gradio app repo/app.py
+++ b/HF DS Gradio app repo/app.py
@@ -1,0 +1,16 @@
+import gradio as gr
+from guidance_helper import generate_html  # Import the function
+
+with gr.Blocks(fill_height=True) as demo:
+    with gr.Sidebar():
+        gr.Markdown("# Inference Provider")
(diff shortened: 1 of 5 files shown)


jashonnew:
There were some resolution bugs requiring me to run parts of the deepseek model in order for guidance to have full access. While I want to do this in the future, it seemed beyond scope for this assignment to dive into the code. I wrote tests to use in my Huggingface Space. Obviously it is never fully satisfying when you do not see the output you desire, but I feel much more confident on the topic of constrained output and feel as though I could use guidance if required in a work/school setting. Maybe I will return here when I am completing my final project.

PR #73, opened 2025-03-28
HW assignment 9

rj-may:
This assignment seems fairly straightforward (that means it is probably not).  
The first step  will be  to load a model. I hope to do that using probably llama 3.1 8B.

Then, I think I will use guidelines. That is in the book and based off some conversing with ChatGPT  it has some advantages over outlines. It appears to work welll with python and I don't want to deal with JSON.

Maybe I will try and make it output into either QA  or maybe some joke/riddle format. 

From there that should be done, but obviously easier said then done.


rj-may:
Let's make a list of things that have gone wrong.  This was a suggestion from chat gpt. 
```
program = guidance("""
<html>
  <head><title>{{gen 'title'}}</title></head>
  <body>
    <h1>{{gen 'heading'}}</h1>
    <p>{{gen 'body'}}</p>
  </body>
</html>
""")
```
TypeError: module, class, method, function, traceback, frame, or code object was expected, got str



Then  I thought huh, the example on the docs uses gguf format. However, I can't convert the llama model to gguf format according to some reading. So that sucks 



Now on attempt c is for cookies I tried to something else and load it directly, but I am running into ram issues.

rj-may:
This assignment took 4 iterations of doing different things. They are named with names a, b, c, and d. The final one assn9_dumb. 
I had some issues, so I spent a long time trying different things. I tried loading llama. 3.1 8b but that didn't really work with guidance. Then I tried a few different things.  I then ran into memory issues. 

I ultimately had to simplify things and I created something that followed the book. I am not even sure if that is what was wanted in terms of output and html format.   

I wish we had some more direction on this assignment. It said to use a big model, the big model I used caused memory issues. I didn't know what format the html should look like, (ex chatbot). I kept it pretty generic. 

I didn't really have time to write tests with this assignment. I have  a lot of homework happening, and I am defending my thesis on April 10.  Busy busy. I also got roped into volunteering at an event tonight. 




Screenshot_demo

PR #74, opened 2025-03-28
HaozheChen Assingment 9

chenhaozhe1:
My plan is to create a Flask web service that takes a prompt via a POST request and returns an HTML page with the model’s generated response, using the Falcon model through the guidance library. I’ll wrap the output in basic HTML and send it back to the user. I’ll ensure the service handles errors properly, such as when no prompt is provided or if something goes wrong. Afterward, I’ll test it with different prompts to make sure everything works smoothly.


chenhaozhe1:
Everything went pretty smoothly, and I was able to set up the Flask app and tests as planned. The tests covered key functionality like handling valid and invalid prompts, checking the HTML response, and ensuring the model behaves correctly. I wrote the tests to ensure that the Flask service works as expected and handles errors properly. If I had more time, I’d add more edge cases and tests for model-specific behavior, and I’d also improve the error handling to cover more scenarios. Overall, it’s in good shape!

msharp9:
Just a note, this isn't actually using constrained sampling, albeit you are using guidance for generation.  Well, technically you did what the homework assignment asked by generating HTML output.  Maybe something for you to come back to and explore more later.

PR #75, opened 2025-03-28
Assignment 9 - Structured Output

m6urns:
I've decided to implement structured HTML output using the Guidance library with OpenAI's API integration. This approach provides explicit control over output structure through templates hopefully without complex local setup. I considered Outlines as an alternative, but after reviewing both options it seemed that Guidance would be easier to use for structured HTML output. 

I'll create a Flask application with a single endpoint that processes POST requests, passing parameters to a Guidance program that fills an HTML template using OpenAI's model. The service will include error handling for invalid inputs, API failures, and sanitization of the generated HTML to prevent security issues.

Testing will verify valid HTML output, input handling, error management, and structure consistency in test_guidance.py.


m6urns:
I implemented a structured HTML output service using the guidance library with OpenAI integration. The guidance library worked well for this use case, it's template-based system with chat model support allowed us to enforce proper HTML structure without complex post-processing.

I created a Flask application with a REST API endpoint that accepts POST requests containing a topic and content type. The service processes these parameters, validates inputs, and generates appropriate HTML based on the content type. For article pages, the system creates a structured document with title, sections, and footer elements. For product pages, it includes features lists, pricing, and buy buttons, ensuring each follows proper HTML semantics.

For testing, I created a test suite using pytest that validates both the HTML structure generation and error handling paths. API-dependent tests are marked to be skipped when no API key is available, ensuring tests can run in various environments. 

Screenshot from 2025-03-28 18-06-20

PR #76, opened 2025-03-29
Matthew white assignment 9

matthewwhite1:
Plan of action:

- Look into outlines and guidance
- This might be slightly off from what you wanted us to do, but I am deciding to use outlines to generate JSON output, and then convert that output to HTML using the json2html module
- I will set up a file very similar to the JSON generation file in outlines
- I will ask ChatGPT to help write tests for the code


matthewwhite1:
Here is the output of my constrained output function before I added ZenML stuff, given the prompt "Give me an old character that is incredibly strong":
Screenshot 2025-03-28 193626

matthewwhite1:
For once, this was actually easier than I thought it would be, but that may be because I misunderstood the assignment and/or simplified it too much. I did end up directly using code from the outlines README to constrain the output to JSON, and then I converted that output to HTML. ChatGPT wrote the tests for the code. I tried a few things to fix the GitHub checks, but I eventually gave up after a while.

msharp9:
Good job, and smart move. I'm doing essentially the same thing at work right now. Using constrained sampling to generate JSON I could then translate to the actual format I wanted.

PR #77, opened 2025-03-29
Erick graham assignment 9

Chazzwazzler:
I will be using a Llama model for this assignment. If possible I will just do the one I have been using for the past few assignments, but if I need to switch to a larger model, I will.

I will be using Guidance to structure the output of the model into HTML. The documentation for Guidance seems to be pretty good and it doesn't output JSON, unlike Outlines. I will write tests to detect if it generates broken HTML, and other general stuff like if it generates errors and such.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,3 +1,24 @@
-def hello_world(name: str) -> str:
-    return f"Hello, {name}!"
-  
+from guidance import models, gen
+from guidance import user, assistant, system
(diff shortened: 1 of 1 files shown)


Chazzwazzler:
image

msharp9:
Looks good! Nice and simple.  Missing tests and explain your work portions.

PR #78, opened 2025-03-29
Homework assignment 9

tjkerby:
My plan for this assignment is to use outlines for the structured output. I chose this over guidance mainly because it seems like outlines plays nicer with langchain and I'm hoping to leverage langchain/langgraph for my final project. As far as models to use, I plan to use the newly released gemma-3-12b-it model and see how it works. If I have issues I will likely switch to the the llama3.3 8 b model or possibly a similarly sized deepseek model. As far as tests go I will run tests to check that the output is valid html.


tjkerby:
image
I had a hard time getting the new google gemma model to work and ended up switching to the llama3.1 model. I had a hard time figuring out how to get it to work directly with html and found that generating json first and then converting to html worked out better for me. The first test shows that it is a valid html file by running beautiful soups html parser on it. The second test make sure that it has the proper fields with the proper outputs. I'm not sure exactly why the tests are failing when it is on github, but they work on my machine.
image

PR #79, opened 2025-03-29
Homework Assignment 9

eliasthompson304:
PLAN:

So I assume this project will be a lot like last project in that I serve my model. The difference in that the output is to be HTML formatted rather than just a text output on the terminal. In order to accomplish this, I will:

1) Learn how to format my output in app.py in html. 
2) Understand how I can view this output on my web browser
3) Run tests to ensure the HTML formatting is working properly


eliasthompson304:
My scripts host a local server using a FastAPI and give an HTML structured output. If I had more time I would have gone back to my training script and maybe tweaked a few things, like the learning rate of number of epochs. I also might have grown my dataset even more before re-training. I wrote the test I did to ensure the HTML structured output came as expected. This is what I see on my end: 
Screenshot 2025-03-28 at 9 57 50 PM

I still need to work on cohesiveness but it looks better than last week's results (the word "the" printed hundreds of times). To fix this issue of repetitiveness I only needed to tweak my app.py script and add some additional parameters to the model.generate() method. I gave it a penalty for repetitiveness, a temperature of 0.7, set a top_k param to 50(which reminded me of when I used Pinecone) and also top_p to 0.9 (which I learned dynamically adjusts the number of words considered for each prediction based on the probability distribution, rather than fishing through the entire list of possibilities each time).

eliasthompson304:
So now it seems the only "check failed" is that I do not have access to my own repo on huggingface. I am logged in on huggingface using huggingface-cli login, so ig I will leave it as is.

msharp9:
This is really good work! Love the screenshot showing the html rendering.  Good job prompt tuning to get better results.  Don't worry about the huggingface access on your tests, they look good.

PR #80, opened 2025-03-29
Original plan

Nathan-Weber208:
# Plans for assignment 9

Obviously the easiest method would be to take the existing code i did last week and just change the fastAPI server to put it in a p tag. Poof! HTML!

But since I want to actually learn something ill give the guidance example a try.
I will try to use the model to generate html and ill push it to the website and display it there.


Nathan-Weber208:
I diddnt end up using guidance. it was far easier to just manually change the html to fit. 

If i had more time, i want to check out the options for a better frontend. Streamlit is something a friend of mine recommended. you can expect a future commit of mine including it. I like the idea of my work being similar to the pros rather than just a flask endpoint with a vibe coded frontend.

Nathan-Weber208:
its easy to run BTW its all in the dockerfile. you just need your hugging face token

msharp9:
If you are looking for frontend options Chapter 8 of the book goes over several options like Gradio to Streamlit.

PR #81, opened 2025-04-04
HaozheChen Assignment_10

chenhaozhe1:
For this assignment, my plan is to use DSPy with the gpt-4o-mini model and build a simple agent that decides what to do based on the user’s input. I’ll set up a router module that generates a natural language explanation of what it wants to do. Based on that, the agent will either go down the internet route, where it generates a search query and answers the user’s question using mock web results, or it’ll take the prompt and try to improve it using a prompt optimizer. The routing is entirely generative, so the logic flows naturally from the model’s output.


chenhaozhe1:
Here are my submissions:

agent.py:  code of prompt optimizer
Implementation.ipynb: notebook of implementation
test_agent.py: test 

Things mostly went as planned, but I did run into some issues with getting the language model to load properly due to authentication and quota limits. Since the assignment didn’t require actual model output and emphasized avoiding .select or .choice, I decided to focus on testing the structure and logic of the agent itself. I wrote a few simple tests just to check that the modules return something sensible — for example, making sure the optimizer returns a string, the internet agent outputs the mock result, and the conditional agent routes correctly when I mock the decision step. If I had more time, I’d add better mocking for the LM and test a few more edge cases, but overall I was able to keep things clean and modular without needing to rely on a real model.

PR #83, opened 2025-04-04
initial commit hw10 - POA

jashonnew:
I have read the assignment as well as the assigned materials. I am worried I don't quite understand the assignment. It seems like we should be able to use dspy in some way to try and prompt engineer the model into responding in a specific way without forcing it to.

What I am going to try to do -> I am going to try using the kaggle notebook method to load in a model via dspy and then I am going to try and coerce the model via dspy into responding to me via email. I would like all of the return formats to be via email.

I have run into api issues before, if that is the case and for that reason its not working then I may try getting it to work via a hugging face app so that way I am directly connected to the model.

### DIFF
--- a/dspy.example.py
+++ b/dspy.example.py
@@ -1,25 +1,36 @@
 import dspy
-from dspy.datasets.gsm8k import GSM8K, gsm8k_metric
 
-# Configure the dspy settings
-dspy.settings.configure(lm=dspy.LM('ollama_chat/llama3.2:1b', api_base='http://localhost:11434', api_key=''))
 
-# Load the GSM8K dataset and a few examples
-gsm8k = GSM8K()
-gsm8k_trainset = gsm8k.train[:10]
 
-# Define the dspy program and optimizer
-dspy_program = dspy.ChainOfThought("question -> answer")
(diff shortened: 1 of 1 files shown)


jashonnew:
I would like to get this to actually work, I feel within the environment that we are working now, which is a local app with an llm api key, I would really like a demo on setting that up. How are we going from a place with no model chosen and no api key selected into having it locally available to us. I haven't been able to make that happen yet and at this point I really could use some help. 

I got two different examples I was trying, one in kaggle, one within the HuggingFace Spaces framework. I underestand them theoretically and believe that if I could resolve the api issues they would in fact work (one returning floating points, one returning responses that were always formatted as an email). 

While I don't actually have tests written, If I could visualize the actual output of the two different models I attempted this would inform what I would like the output to look like and then I could format some tests that would check for and help me ensure the selected output type.

msharp9:
It seems like maybe you need a crash course in APIs?  Happy to set up an office hours or feel free to ask this at the end of class on Monday. Happy to stay late and answer all your questions.

msharp9:
Don't forget to write tests next time.

PR #84, opened 2025-04-04
Nathan weber 10

Nathan-Weber208:
First I'm going to upgrade the loaded llm with something more substantial. 

then, I'm going to use dspy to pull data from the internet on specific queries.

this will be similar to the agent i made in the past except it will be pulling from something more substantial than a simple list of inputs and outputs.


Nathan-Weber208:
Check out the readme for a photo of the result.

I tried with a more complicated model, but it took too long to run. im talking over an hour. so i went back to last weeks model.

But it turns out wikipedia has a super convenient API that allows you to search using it. so i used that as a simple agent.

If i had more time i could have made the model only reach out to Wikipedia if it was unsure of the answer.

msharp9:
Don't forget to write tests next time.

PR #85, opened 2025-04-05
Erick graham 10

Chazzwazzler:
I'll probably just use Llama 3.2-1B again for this assignment. It's worked fairly well for me so far and I already have the .gguf file made for it. 

I'll use DSPy to get information from Wikipedia when necessary and feed that into the model. It should be fairly simple to pull off, DSPy seems to handle most of the work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Install `uv` using the instructions found here: <https://docs.astral.sh/uv/getti
 
 If you need to add dependencies to run your code, for example cowsay, you can do so with:
 
-```bash
+```bash 
 uv add cowsay
 ```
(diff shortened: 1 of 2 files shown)


Chazzwazzler:
DSPy is working, though I have been having some difficulty getting it to pull up to date information from the internet. From my testing it is actually getting information from Wikipedia and entering it as context to the prompt, though.

image

image

msharp9:
It looks like you forgot to push your code. It looks like you did it though, so if you can push the code before class on Monday I'll give you a pass being late.  Otherwise I can only give you points on what I can see.

PR #86, opened 2025-04-05
Assignment 10 - Matt Burns

m6urns:
I will modify my code from Assignment 9 to accomplish this task. I am going to modify my approach and use more sophisticated prompt engineering than what was used previously to accomplish structured output without constrained sampling.  I will modify my existing flask app and add an additional Prompt class. I will also implement unit tests to validate my approach.


m6urns:
This assignment was fairly straight forward and by modifying my previous assignment to use more specific prompting to accomplish structured output in the PromptOptimizer class went well. I was able to implement non-constrained prompt engineering using detailed and specific instructions in my prompt as well as few-shot prompting.  Additionally I was able to establish context with the LLM model using system prompting. I didn't mention in my initial plan but ended up needing to include a new structured-output endpoint. Additionally, I was able to implement unit tests to test the functionality of my implementation. 

Screenshot from 2025-04-04 19-54-18

PR #87, opened 2025-04-05
Matthew white assignment 10

matthewwhite1:
Plan of action:

- Figure out what this assignment is even asking for
- Try to run the dspy code from class
- Choose the "optimize a prompt" option on the assignment because that sounds easier
- Ask Llama one of the tricky LLM questions and mess around with prompts until I find one that works


matthewwhite1:
image

matthewwhite1:
I chose to prompt engineer a prompt to get Llama to answer a tricky question.

The tricky prompt I used was taken from here:

https://matchingpennies.com/hard_questions_for_llms/

I used the prompt: "If I hang 5 shirts outside and it takes them 5 hours to dry, how long would it take to dry 30 shirts? THE CORRECT ANSWER IS NOT 150 HOURS."

I tried many different strategies to get Llama to successfully answer this, but unfortunately, none of them could correctly get the answer of 5 hours, and I ran out of time.

I tried adding the orchestra question as an example, but it ignored that or combined it into its answer to the shirt question, with something like "it would take 60 hours for 600 people to hang 30 shirts," which was funny. I tried saying, "DO NOT USE ANY MATH TO ANSWER THIS QUESTION." But, since it is trained on math questions, I don't think it could resist the urge to do math anyway. I also tried to restrict it from giving certain answers like 150 hours, but it would either make even worse answers or ignore the prompt.

I am honestly not sure if this small version of Llama trained on these few math questions has the capability to answer this question.

PR #88, opened 2025-04-05
Homework Assignment 10

eliasthompson304:
Plan. -

I intend to build an agent that can use the internet to give relevant and updated information. In order to do this I will:

Use the DDGS from the duckduckgo_search package (others were recommended but duckduckgo was the first example I found).

This agent will hopefully have the ability search the web and use what it finds to generate better responses.
I've also wanted to create an agent that can remember prior outputs for my story generating model. I will try to implement that as well. From what I've researched keeping track of session id numbers can be beneficial.

Because my model that I've attempted to fine tune was for story generation, I will use a different model for this assignment. I'd rather just experiment with an untouched model meant to give accurate responses.


eliasthompson304:
I started out by trying to create a simple agent that would remember things. Soon I realized that my fine-tuned model generated nonsense anyway, so it would be difficult to tell whether the agent was truly making a difference. Due to the fact that I was going to implement an agent that uses the internet and extracts factual information, I decided to just switch out my model choice to a GPT2 model. From there I built up my agent using duckduckgo and some help from AI. I realized that my prompts were not cutting it, so now I use a method that structures an optimized prompt for any input the user might give. Here is an example of it running:

Screenshot 2025-04-04 at 10 36 13 PM

On the right screen you can see the web results the agent found, and on the left is the subsequent response. Still not quite there, but the web is being accessed and hopefully influencing the model's output.

This brings me to the tests. My first test is there to check whether the agent actually was fetching web info. My second test checks whether when the memory (attribute of the agent) is updated, whether that update really happens and stores the correct user input and model output.

PR #90, opened 2025-04-11
done with plan

rj-may:
Assignment 11 game plan

Step 1. Download  Ollama and and small llama model.

Step 2. Pip install gradio. 

Step  3. Make simple script that talks to the llama model. 

Step 4.  talk to the llama via an api request.

Step 5. Repeat steps 3 and 4 till it works. After many hours of this settle for whatever I get out.


rj-may:
HW11_screenshot


I feel that this assignment went pretty well. This is the assignment were plan actually worked out. I was able to get gradio up and running. I think I may have had more success with this assignment because I didn't limit myself to collab. 

The way installed Ollama was on Windows and my environment with WSL2 is where I had pytests installed. I didn't really know how to test this because it seemed to be one of those where either it ran or it didn't. I had chatgpt write some tests that seemed to work.

PR #91, opened 2025-04-11
initial commit hw 11 frontend - POA

jashonnew:
In preparing for this week's assignment, I already have some familiarity with Gradio because of my work the last several weeks within the Hugging face environment. I am going to continue to use Gradio and hopefully fix my issue with APIs. I watched several videos on APIs and am going to invest some time this week into discovering how to get a use-able API for an llm for employment in my final project as well. I think this investment will rocket my learning on this topic. 

I think if I will struggle anywhere it will be in running the application locally vs. hosted on Hugging face via the huggingface spaces . The ease of integration in that environment with gradio is really intuitive. The translation of that interface into a local environment will be the difficult part. I will also be sure to run ruff as well as write tests for whatever I end up completing this week .


jashonnew:
This went great! I am stoked with my finished product. There seems to be some history repetition that I would like to clean up but overall I am very happy with what I was able to accomplish. I ran ruff to format my code, and wrote a test file gradio_tests. 
app_working

In the end I figured out using ollama to host a local model of the llama 3.1B model and using the terminal i had a live version of the model and used a gradio app to interface it.

PR #92, opened 2025-04-11
Tom kerby 11

tjkerby:
My plan here is to use gradio to create my frontend for my llm application as I'm interested in being able to build dashboards for other models that might have image or audio inputs (which from my brief research I've learned gradio works better with). As far as tests go, since I'm using ollama, I will test that ollama can find and load the model, that it can receive a request and return a response. I will also make sure that it can handle multiple responses.


tjkerby:
image

This was one of the more fun assignments for me. I'm honestly surprised that it wasn't harder to get this up and running. I had enough fun that I figured out how to get persistence using a SQL database, and give it the option to pick different models that I already have loaded using ollama. I tested that streaming works, that the chat successfully updates, among a few others.

Overall, this makes me really want to figure out how to get this to work with images because if I can do that it would be much easier for uploading images for local models.

PR #93, opened 2025-04-11
Nathan weber 11

Nathan-Weber208:
My plan is to connect streamlit as a frontend.


Nathan-Weber208:
This ran very smoothly. It went exactly how i planned.
I wrote the test that i did because i was just testing the frontend. so I made the Wikipedia part just be dummy response.

If i had more budget id like to make a real beast of a token processor. I really want to be able to get something working locally that thinks as fast as chatgpt. or at least something faster than a question every minute.

msharp9:
While technically that is a plan, it doesn't give me much to understand your thought process to evaluate it.

PR #94, opened 2025-04-11
Matt burns Assignment 11

m6urns:
My plan for this project is fairly simple, at the end I intend to implement a simple CLI interface for my current codebase. I would like to implement this interface in the CLI and would like to be able to stream generic responses from it. I will utilize my existing PromptOptimizer class and to implement the interface I will use some examples published by OpenAI that provide information on streaming via ChatGPT api calls.


m6urns:
Completing this assignment was fairly simple, and putting together a simple streaming CLI interface for my existing codebase went well. I think that if I was to put further effort in to this I would allow users to provide more specific system prompts and to include their previous prompts in the context window. 
Screenshot from 2025-04-11 16-24-36

msharp9:
Great work, looks like you are missing tests for your CLI.

PR #96, opened 2025-04-12
Homework Assignment 11

eliasthompson304:
Plan:

Read the Gradio documentation and then use to create an intuitive front end that users can interact with. From what I've read and heard it seems to be a very ideal way to present your LLM to a less technical audience. My first peak at the documentation also gives me the feeling that it won't be too impossible.

I will use a llama 3.2-8B foundation model this time just to experiment with a new model. So far I've only used a 1B model and a foundation GPT2 model since working on local servers and agents. I plan on keeping my app and agent scripts from the last assignment.
My goal is simply to create a simple "question" and "response" interface with a nice submit button people can press to test out the LLM. 

I'm having a hard time thinking of any tests at the moment. I will think of some soon hopefully.


eliasthompson304:
After reading the gradio documentation I was able to take the web agent I built last week and alter it's output from merely being on the command line to something intuitive and better-looking. Gradio's documentation is fortunately not hard to follow, and the basics are now appleid to my app.py script. I tried using a llama 8B model but due to some unforeseen complications I chose to just stick to the same model as last time. Here is an example of Gradio in action: 

Screenshot 2025-04-11 at 10 04 56 PM

As of now it is kind of hit or miss. Sometimes it gives relevant and accurate responses with some random sentences here and there. Something the web agent doesn't do anything and I get a one sentence response indicating a confused language model. If I had more time to improve this pr I would use a larger model and learn how to make my web agent more consistent and yield more accurate results. I might have also worked to create a prettier UI. I chose to stick with the same tests as last time, to be honest I couldn't think of any great test cases for a front-end.

msharp9:
it looks like you just renamed your agent tests file but aren't actually testing your gradio front end at all.

